PR #2, opened 2021-07-23
Render templates

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4948

This affects the way the plugin displays the social icons, extra listing images, buttons, and pagination.

We should test these on the single listing page and all listings with the default theme, no theme, and maybe a couple other themes as well.

It would also be best to test these shortcodes:
https://businessdirectoryplugin.com/knowledge-base/shortcodes/#businessdirectory-images

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -13,6 +13,7 @@ node_modules export-ignore
 package-lock.json export-ignore
 package.json export-ignore
 phpcs.xml export-ignore
+phpmd.xml export-ignore
 phpunit.xml export-ignore
 README.md export-ignore
 tests export-ignore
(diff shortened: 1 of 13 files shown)


garretlaxton (commented):
I was stuck for a minute on this one, but I got the part templates to override by placing them in `<active theme>/business-directory/parts/`. The docs don't mention the parts folder, I can add that if you'd like.

The social icons, extra images, buttons, and pagination all look fine as well as the themes.

I just couldn't get those shortcodes to display anything, with and without using an `id`.

stephywells:
Thanks @garretlaxton ! I should have mentioned the parts folder. I should have thought of that, since I got held up on it too because I was trying <active theme>/parts/. Sorry about that!

PR #3, opened 2021-07-28
Issue 4950

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4950

The submit listing page should now be overridable.

I need to look into the difference between these two function more before merging. I don't understand why these would be used in different cases.

### DIFF
--- a/includes/admin/class-themes-admin.php
+++ b/includes/admin/class-themes-admin.php
@@ -76,6 +76,8 @@ class WPBDP_Themes_Admin {
     }
 
     function pre_themes_templates_warning() {
+		global $wpbdp;
+
         $pre_themes_templates = array(
 			'businessdirectory-excerpt',
 			'businessdirectory-listing',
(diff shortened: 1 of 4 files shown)


stephywells:
@garretlaxton This should be better now. wp_render and wpbdp_x_part will both check the same places for template files.

The thing is, some template files are specifically excluded from getting overridden. That includes the submit-listing-images-upload-form and submit-listing templates. The thing is, that this can get too difficult to maintain if we let people override everything. Instead, I think we'll need more info on what people are trying to override in any templates that aren't currently allowed. Then we can take it on a case-by-case basis.

garretlaxton (approved):
Should we mention which templates aren't meant to be overridden in the docs? It may also be helpful if we get a ticket, we will know if that file can be overridden or not.

stephywells:
It might be easier to list the files that can be overridden with a note that anything else most likely can't be overridden. Do you have more detail on what people were wanting to customize in the submit listing templates?

garretlaxton:
@stephywells That will work too. The user was trying to change the order of the submit listing process.

stephywells:
@garretlaxton Thanks for the extra detail. That isn't actually something that would be changed in the template. It would need to be done using a hook, or it would break things.

PR #6, opened 2021-07-30
add support for using id=# in a few shortcodes

stephywells:
These shortcodes will now work with the id:
[businessdirectory-socials id=251]
[businessdirectory-buttons id=251]
[businessdirectory-details id=251]
[businessdirectory-images id=251]

### DIFF
--- a/includes/class-shortcodes.php
+++ b/includes/class-shortcodes.php
@@ -571,7 +571,7 @@ class WPBDP__Shortcodes {
 			return '';
 		}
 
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
These shortcodes are working perfectly with an ID! I used the Safeway listing for the ID. 

shortcode-ids

When using these with Elementor, I'm guessing the pro version required? I believe that's where I was having trouble yesterday as I only have the free version.

stephywells:
Awesome! Thank you!

Yep, the templates are only in Pro.

PR #7, opened 2021-08-03
Fix issue with checking out with 100% off code

stephywells:
https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4949

With 100% off, this hides the billing and credit card form. It also prevents sending the payment to the gateway.


garretlaxton:
Listings now go through when using a 100% off discount. Nice!

Screenshot 2021-08-03 at 12-41-55 Business Directory – Just another WordPress site

stephywells:
Awesome! Thanks @garretlaxton ! It looks like there are more issues we need to fix, but there are separate issues logged. I didn't realize coupons don't even work with Stripe.

PR #8, opened 2021-08-23
Remove Ajax compatibility mode

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4954

Ajax compatibility mode deactivated all non-BD plugins when an ajax request was made. This was designed as a workaround for plugin/theme conflicts when they output whitespace before the page is loaded. This is common when a php file ends with "?>
(This is a blank line)
"

It looks like WordPress handles this now, because I'm not seeing issues with the whitespace now. Going backward pages in the listing form, and clicking the "change plan" link are good examples to check. Those are the most critical.


stephywells:
I think WordPress must have done something to prevent this from being an issue... It fails if there's an error message or content where the white space is. Hopefully that doesn't cause problems on many sites.

PR #17, opened 2021-08-30
Directory structure update: Assets Folder

theRealRizeo:
- Move asset vendors (jquery-breakpoints, jQuery-File-Upload and selectWoo) from  the vendors directory to the assets/vendor directory. This makes it more consistent that all assets are in one directory. 
- Add constant `WPBDP_ASSETS_URL` to be used for asset path for consistency within the entire plugin.
- Update assets path to use the constant `WPBDP_ASSETS_URL` when referencing urls for assets


Gruntfile.js
```diff
@@ -81,15 +81,15 @@ module.exports = function( grunt ) {
       if ( ! _.isEmpty( less_config ) ) {
         grunt.config.set( 'less.' + id, {options: {cleancss: false, compress: true, strictImports: true}, files: less_config} );
         grunt.config.set( 'watch.' + id + '_less', {
-          files: [path.join(basedir, '**/*.less'), path.join(basedir, '**/*.css'), '!' + path.join(basedir, 'vendors/**/*'), '!' + path.join(basedir, '**/*.min.css')],
+          files: [path.join(basedir, '**/*.less'), path.join(basedir, '**/*.css'), '!' + path.join(basedir, 'vendors/**/*'), '!' + path.join(basedir, '**/*.min.css'), '!' + path.join(basedir, 'assets/vendor/**/*')],
```
stephywells, line 84: We need these js files in the packaged zip. Let's revert this change to make sure we don't lose these files in builds.

includes/class-assets.php
```diff
@@ -57,36 +57,36 @@ public function register_common_scripts() {
...
-		wp_register_style( 'wpbdp-base-css', WPBDP_URL . 'assets/css/wpbdp.min.css', array(), WPBDP_VERSION );
+		wp_register_style( 'wpbdp-base-css', WPBDP_ASSETS_URL . 'css/wpbdp.min.css', array(), WPBDP_VERSION );
     }
 
 	private function register_select2() {
         // Use Select2 styles and scripts from selectWoo https://woocommerce.wordpress.com/2017/08/08/selectwoo-an-accessible-replacement-for-select2/.
         wp_register_style(
             'wpbdp-js-select2-css',
-            WPBDP_URL . 'vendors/selectWoo/css/selectWoo.min.css',
+            WPBDP_ASSETS_URL . 'selectWoo/css/selectWoo.min.css',
```
stephywells, line 82: This path isn't right.

stephywells (changes requested):
I'm seeing a couple issues, and then we can merge this. Thanks Paul!!

Gruntfile.js
theRealRizeo: This tells grunt not to watch this directory for any changes. The vendor directory should not be included in the path. The same items in the `assets/vendor` directory where in the `vendors`, I have just appended the grunt task to exclude this also

stephywells: You're right. Thanks!

stephywells (approved):
Looks good now!

theRealRizeo:
Hey @garretlaxton just a reminder on the review of this :)

garretlaxton (approved):
@theRealRizeo Whoops! I thought this was something else, sorry about that. Thanks for the reminder!

Changes look a lot cleaner. Everything still functions as should.

PR #18, opened 2021-08-31
Abstract Directory

theRealRizeo:
- Introduce the abstract directory
- Move abstract `WPBDP_Form_Field_Type`  class and abstract `WPBDP__Payment_Gateway` to the `includes\abstracts` directory

After this update, will move other files


stephywells (changes requested):
@theRealRizeo let's add the bit of reverse compatibility we mentioned. Can you add those two files back? They'll just look something like this:

```
_deprecated_file( basename( __FILE__ ), '5.14', null, 'This file can be found in includes/abstracts/class-form-field-type.php' );
require_once WPBDP_INC . 'abstracts/class-form-field-type.php';
```

stephywells (approved):
Thanks so much! This looks perfect!

PR #19, opened 2021-09-01
Back button on add listing form malfunctioning #4951

theRealRizeo:
This is related to https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4951

The back button would show on default when there is no previous section, yet the back button is only meant to show when there is something to go back to


garretlaxton (approved):
Looks good! No reason to have a back button for step 1. Not sure why I never ever noticed it myself... haha

Screenshot 2021-09-01 at 14-41-59 Add Listing – BD Demo

stephywells (approved):
Nice! Thanks so much @theRealRizeo ! I don't think I've ever noticed that either. Weird!

PR #20, opened 2021-09-01
Feature : Listing contact phone field

theRealRizeo:
This adds a mandatory phone field to the contact form of the listing. The field is also inserted into the email sent out to the listing owner


includes/views/listing_contact.php
```diff
@@ -55,6 +57,10 @@ private function validate() {
             $this->errors[] = _x( 'Please enter a valid email.', 'contact-message', 'business-directory-plugin' );
         }
 
+        if ( ! $this->phone ) {
+            $this->errors[] = _x( 'Please enter your phone number.', 'contact-message', 'business-directory-plugin' );
```
stephywells, line 61: Let's keep the phone field optional. Maybe we can include a hook for people to add custom validation instead?

stephywells (commented):
Awesome! Thanks so much for adding this!

includes/views/listing_contact.php
theRealRizeo: @stephywells I have pushed an update that includes a filter. Should the validation message also be custom or should we leave it to the filter `wpbdp_contact_form_validation_errors` to handle this?

stephywells: I was thinking we could add a filter on the $this->errors array so anything could be added. We can add a code example on our site to make the phone required.

theRealRizeo: @stephywells there is documentation of this here https://businessdirectoryplugin.com/knowledge-base/add-phone-number-to-listing-contact-form/ on how to add custom validation. Though I am not sure how users will take to having to use a filter to add validation messages for a file that comes with the plugin

stephywells: Oh, right. There's already a hook for this.

I think generally, best practice is to not require phone fields for things like this. The example here would cover this great:
```
function example_validate_custom_contact_field( $errors ) {
  if ( empty( $_POST['phone-number'] ) ) {
    $errors['phone-number'] = 'Phone number is required!';
  }
  return $errors;
}
add_filter( 'wpbdp_contact_form_validation_errors', 'example_validate_custom_contact_field' );
```

Ideally at some future point, we'll have an easier way to replace this with a Formidable form of something, so it can be completely customized.

templates/listing-contactform.tpl.php
```diff
@@ -34,6 +34,10 @@
 		</p>
     <?php endif; ?>
 
+    <p class="wpbdp-form-field">
+        <label for="wpbdp-contact-form-phone"><?php esc_html_e( 'Phone Number', 'business-directory-plugin' ); ?></label> <input id="wpbdp-contact-form-phone" type="text" class="intextbox" name="commentauthorphone" value="<?php echo esc_attr( wpbdp_get_var( array( 'param' => 'commentauthorphone' ), 'post' ) ); ?>" />
```
stephywells, line 38: Let's us "tel" here for the input type.

includes/views/listing_contact.php
```diff
@@ -55,6 +57,18 @@ private function validate() {
...
+        /**
+         * Optional phone validation.
+         *
+         * @param bool $validate - defaults to false to skip phone validation.
+         * @param string $phone - the phone field.
+         *
+         * @return bool
+         */
+        if ( apply_filters( 'wpbdp_contact_form_validate_phone', false, $this->phone ) ) {
+            $this->errors[] = _x( 'Please enter a valid phone number.', 'contact-message', 'business-directory-plugin' );
```
stephywells, line 69: Oh, let's add format validation here when the field isn't blank. Here's the regex we're using in Formidable:
https://github.com/Strategy11/formidable-forms/blob/a6b06bcc7a0fe6a3ab5cdc0ea848f79697b960ad/classes/models/FrmEntryValidate.php#L177

theRealRizeo: Sounds good, I know where to put it :)

theRealRizeo:
@stephywells I have added phone validation in the validation class. This will also allow us to extend the validation to the pone field.
I have also fixed the CSS of the `tel` field type to take the 100% width as the other elements

garretlaxton (approved):
The new phone number field looks good and functions as should! Nice new addition!

Screenshot 2021-09-01 at 16-37-40 Andy Agent – BD Demo

includes/form-fields.php
```diff
@@ -585,6 +585,7 @@ public function get_validators() {
                 'decimal_number' => __( 'Decimal Number Validator', 'business-directory-plugin' ),
                 'date_'          => __( 'Date Validator', 'business-directory-plugin' ),
                 'word_number'    => __( 'Word Count Validator', 'business-directory-plugin' ),
+                'tel'            => __( 'Telephon Number Validator', 'business-directory-plugin' ),
```
stephywells, line 588: There's a typo here.

stephywells (commented):
Nice! Just one last change and I think it's ready to merge.

includes/form-fields.php
theRealRizeo: Fixes

stephywells (approved):
Looks awesome!

PR #23, opened 2021-09-09
Listing statistics feature

theRealRizeo:
Adds listing statistics.

- [ ] New admin setting, a sub-tab under the `Listings` tab called `Statistics` `/wp-admin/admin.php?page=wpbdp_settings&tab=listings&subtab=listings%2Fstats`
- [ ] Option to turn `off` or `on` and to allow tracking of admins. Defaults to `off` for both
- [ ] New column in Listings called `Views` that will only appear if `Enable statistics?` is checked
- [ ] Statistics saved in a new database table `{database_prefix}_statistics`
- [ ] Views are counted once per user per day
- [ ] Constant check in the Statistics model called `WPBDP_LISTING_VIEW_ENABLE_TRACK_IP` that when defined as false, does not track the IP of users

### DIFF
--- a/includes/admin/class-admin-listings.php
+++ b/includes/admin/class-admin-listings.php
@@ -316,8 +316,11 @@ class WPBDP_Admin_Listings {
         }
     }
 
+    /**
+     * View column
+     */
     public function listing_column_views( $post_id ) {
-        $statistics = new WPBDP_Listing_Statistic();
+        $statistics = WPBDP_Listing_Statistic::instance();
         echo $statistics->count_views( $post_id );
     }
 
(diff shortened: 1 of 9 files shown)


theRealRizeo:
listing-admin-table

admin-setting

garretlaxton (approved):
New view counter works good!

stephywells:
@theRealRizeo Can you move this new feature into BD Premium? From here on out, I don't think we'll be adding features to the free version. If it's a bug fix, it will be in the free version. Otherwise, in most cases it'll be a premium feature.

theRealRizeo:
@stephywells Sure thing :) Also I think the Field Icons PR I will move it :)

PR #24, opened 2021-09-13
Feature : Custom field icon

theRealRizeo:
- This adds a custom field icon for each field that can be managed when editing or creating a new field
- The field icon is displayed when viewing a listing
- Right now supported icons are FontAwesome and DashIcons

NB : Might not work well if previous PR of https://github.com/Strategy11/business-directory-plugin/pull/23 was tested. Mainly because Database version in this has also been incremented by 1


theRealRizeo:
Moving this to the premium version :)

theRealRizeo:
@stephywells would it be better to leave this in core or in premium addon? Issue is there are changes to the template file for fields and also additional fields that need to be saved that do not currently have actions or any other overrides

garretlaxton:
I am getting the following database error. I did deactivate & activate the plugin this time which didn't help.

Screenshot 2021-09-14 at 14-28-11 Form Fields ‹ Business Directory — WordPress

theRealRizeo:
@garretlaxton mainly because the previous PR updated the DB version :( You may need to add this somewhere in the `functions.php` file `update_option( 'wpbdp-db-version', '18.4' );` then refresh the home page once, and remove the code. After the column will be created

garretlaxton:
@theRealRizeo Appreciate it! I just created the columns myself. 

Dashicons look good! Unfortunately, I just get blank boxes for font awesome in the back-end.
Screenshot 2021-09-14 at 15-38-51 Form Fields ‹ Business Directory — WordPress

theRealRizeo:
Hey @garretlaxton sorry about that, should be visible now. Just pushed an update

garretlaxton:
@theRealRizeo Looks good! 

One last little thing I notice is if you choose font awesome and the modal closes, when you open the modal it will show font awesome icons but the select dropdown says it's Dashicons. It seems the select doesn't save the option.

Screenshot 2021-09-14 at 16-04-31 Form Fields ‹ Business Directory — WordPress

theRealRizeo:
@garretlaxton Will work on that. Also this is not the final view of how it will look, especially in the frontend. So if you have some suggestions please let me know :)

garretlaxton:
I don't think it looks too bad on the front-end as of now. Maybe align it with the text a bit more?

Screenshot 2021-09-14 at 15-37-05 Business Directory – Just another WordPress site

stephywells:
> @stephywells would it be better to leave this in core or in premium addon? Issue is there are changes to the template file for fields and also additional fields that need to be saved that do not currently have actions or any other overrides

@theRealRizeo We can add hooks in the lite version that we can use in the premium version. I'm fine with some parts being included in lite if needed, but the setting to enable it should definitely be in premium.

theRealRizeo:
> > @stephywells would it be better to leave this in core or in premium addon? Issue is there are changes to the template file for fields and also additional fields that need to be saved that do not currently have actions or any other overrides
> 
> @theRealRizeo We can add hooks in the lite version that we can use in the premium version. I'm fine with some parts being included in lite if needed, but the setting to enable it should definitely be in premium.

@stephywells will start on the hooks :)

PR #25, opened 2021-09-15
Add action for fields

theRealRizeo:
Add action in admin field setting form.

- `wpbdp_admin_listing_field_section_visibility` allows to override the `Hide this field's label?` option to use icons in the premium version
- `wpbdp_admin_listing_field_after_settings` allows adding settings after the current field settings


stephywells (approved):
This looks good to me!

PR #26, opened 2021-09-15
Hide image from searchable field type in list

theRealRizeo:
Fixes issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4871 . 
Image field type will not be visible in the searchable field types


garretlaxton (approved):
Image fields are no longer an option for the quick search. Looks good!

stephywells (approved):
Thanks Paul!

PR #33, opened 2021-09-16
Hide listing title on content view and not excerpt

theRealRizeo:
This fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4916


garretlaxton (approved):
I wasn't able to replicate this issue before or after the update. So that's good I guess?

stephywells (approved):
This works great for me. Thanks Paul!

PR #37, opened 2021-09-27
Plugin structure update

theRealRizeo:
- Moving of Admin controllers to includes/admin/controllers
- Moving of admin table classes to includes/admin/views/tables


stephywells (changes requested):
Thanks Paul! This is a good start.

The views folder is for files that are primarily HTML. There shouldn't be any classes in the views folder. It looks like those should be helpers or models. Let me know if you have questions about where to put these.

theRealRizeo:
> Thanks Paul! This is a good start.
> 
> The views folder is for files that are primarily HTML. There shouldn't be any classes in the views folder. It looks like those should be helpers or models. Let me know if you have questions about where to put these.

@stephywells the views folder in the `includes/views` contains classes. It would be better for the html code to b moved to the `templates` folder as thats where files that are not enclosed in classes are mainly

theRealRizeo:
`helpers` should be things used throught the plugin that are repetitive like the `WPBDP_Modules_API` is a helper class used for the api. Views should be what is used to render content and html should be in the templates folder

stephywells:
The templates folder is only for the content we want people to be able to override. Since this adds tons of complication, there is already way too much in the templates folder for the admin area. Ideally, the admin area should be completely under our control and not include templates for override.

In Formidable, we add these table classes in the models folder.

theRealRizeo:
@stephywells should the files under `includes\views` also be moved to `includes\helpers\views` as all the classes in these folder extend `WPBDP__View` class? There are files under `templates\admin` that are used in the admin backend.

stephywells:
Hmm... Those all look like controllers to me since they are primarily routing.

theRealRizeo:
I think `includes/controllers/pages` would be better

stephywells:
@theRealRizeo Was this closed intentionally?

theRealRizeo:
@stephywells yes, I will re-submit shortly

theRealRizeo:
@stephywells I've made some views changes for the files in `includes\views`

stephywells (changes requested):
This looks great!

Since we're moving files around anyway, I'd love to have more consistency in the file names. Specifically, it would be nice to rename any of these files without "class-" in the file name, or anything with underscores. So "form-fields.php" would be "class-form-fields.php". And "manage_listings.php" would be "class-manage-listings.php".

Do you think that would cause any extra possible issues?

theRealRizeo:
@stephywells I don't think it will cause issues and I totally agree with you on that. It will make more sense to have the files named that way, since we already have old references to the files, it should not cause any issues

includes/class-dispatcher.php
```diff
@@ -125,25 +125,35 @@ public function _ajax_dispatch() {
...
         foreach ( $this->get_view_locations() as $dir ) {
             foreach ( $filenames as $f ) {
                 $path = wp_normalize_path( WPBDP_FS::join( $dir, $f ) );
 
                 if ( ! file_exists( $path ) )
                     continue;
 
-                $classname = 'WPBDP__Views__' . implode( '_', array_map( 'ucfirst', explode( '_', str_replace( '.php', '', $f ) ) ) );
+                // Views that start with class- in the file name. Addons may not have this hence the condition.
+                if ( substr( $f, 0, 5 ) === 'class' ) {
```
stephywells, line 151: This is from the PHP manual:

> If you only want to determine if a particular needle occurs within haystack, use the faster and less memory intensive function strpos() instead.

includes/class-dispatcher.php
```diff
@@ -125,25 +125,35 @@ public function _ajax_dispatch() {
...
+                    $f = str_replace( '-', '_', $f );
+                    $classname = 'WPBDP__Views_' . implode( '_', array_map( 'ucfirst', explode( '_', str_replace( $find, $replace, $f ) ) ) );
+                } else {
+                    $classname = 'WPBDP__Views__' . implode( '_', array_map( 'ucfirst', explode( '_', str_replace( '.php', '', $f ) ) ) );
```
stephywells, line 155: It looks like we can prevent this line twice like this:

Set $replace = ''; higher instead of $replace = array( '', '' );

```
$find = array( '.php', 'class' );
} else {
$find = '.php';
}
$classname = 'WPBDP__Views_' . implode( '_', array_map( 'ucfirst', explode( '_', str_replace( $find, $replace, $f ) ) ) );
```

stephywells (changes requested):
Looks great!

I've just got a couple small code cleanup suggestions.

includes/class-dispatcher.php
theRealRizeo: There is a difference in the two lines. With the `class` introduction the name would have an extra underscore. Hence the `if/else` is `WPBDP__Views_` and `WPBDP__Views__`

includes/class-dispatcher.php
```diff
@@ -125,25 +125,33 @@ public function _ajax_dispatch() {
...
         foreach ( $this->get_view_locations() as $dir ) {
             foreach ( $filenames as $f ) {
                 $path = wp_normalize_path( WPBDP_FS::join( $dir, $f ) );
 
                 if ( ! file_exists( $path ) )
                     continue;
 
-                $classname = 'WPBDP__Views__' . implode( '_', array_map( 'ucfirst', explode( '_', str_replace( '.php', '', $f ) ) ) );
+                // Views that start with class- in the file name. Addons may not have this hence the condition.
+                if ( strpos( $f, 'class' ) !== false ) {
```
stephywells, line 150: This would be a bit safer here:
`strpos( $f, 'class' ) === 0`

includes/class-dispatcher.php
stephywells: Where does the extra underscore come from?

theRealRizeo: We only remove the `class` prefix and it leaves an extra `-` that is replaced with an extra underscore. Also the file check if different in the updated code

includes/class-dispatcher.php
theRealRizeo: Issue of checking for `0` as per the docs:

```
Returns false if the needle was not found.
This function may return Boolean false, but may also return a non-Boolean value which evaluates to false. Please read the section on Booleans for more information. Use the === operator for testing the return value of this function.
```

Keeping it as false is safer here as it may return something else

stephywells: It returns the position in the string when it's found. In this case, we are only looking for class at the beginning of the string.

includes/class-dispatcher.php
stephywells: So can we use 'class-' for replace so we can remove that extra line?

theRealRizeo: Other addons load their own views using the class, for now we have to keep both conditions in place

stephywells: Let's not mark conversations resolved until it's done.

This is brand new, so it wouldn't be used anywhere with 'class' in the filename yet. Right? Can you give me an example of a time where using 'class-' for the find wouldn't work?

theRealRizeo: I've made a few changes to check for `class-` and it works both for [claim-listings](https://github.com/Strategy11/business-directory-claim-listings/blob/master/views/confirm_listing_claim.php) and the new file structure. I've pushed cleaner code for the checks

stephywells: This looks much better now. Thank you for sticking it out!

stephywells (approved):

includes/class-dispatcher.php
theRealRizeo: A great learning experience :) Thank you for the guidance

PR #39, opened 2021-10-06
Widget update

theRealRizeo:
This covers https://github.com/Strategy11/BusinessDirectoryPlugin/issues/3866

- Added new widget settings 
widget-settings
- News front layouts


theRealRizeo:
@garretlaxton it will only work when the listing has no image. In your case, the listing has an image :)

theRealRizeo:
@garretlaxton did some tests:

- Listing A has image uploaded
- Listing B has no image uploaded

Both showed fine
widget-listing

The only way it would show as your view is if the image is deleted either from the media library or via the file manager. Could you try view the listing directly and check if the image is shown for the `Shouldn't Expire (should renew)` ?

garretlaxton:
@theRealRizeo Yes, I do have the image both in excerpt and detail views.
Screenshot 2021-10-06 at 13-55-23 Shouldn’t Expire (should renew) – Business Directory

theRealRizeo:
@garretlaxton cant replicate. On my end if I view a listing with no image, there is no image shown as your screenshot. Are you using another addon, custom field, or specific setting?

theRealRizeo:
Oh, okay

stephywells (changes requested):
I'm getting a php error when I try to open the settings for a new widget:

PHP Warning:  Undefined array key "title" in /Users/Steve/Sites/wordpress2/wp-content/plugins/business-directory-plugin/includes/widgets/class-listings-widget.php on line 78
PHP Warning:  Undefined array key "number_of_listings" in /Users/Steve/Sites/wordpress2/wp-content/plugins/business-directory-plugin/includes/widgets/class-listings-widget.php on line 79

stephywells:
@theRealRizeo that didn't solve the problem. Can we revert that and fix it in a more simply way using the existing default settings array?

includes/views/widget/widget-settings.php
```diff
@@ -0,0 +1,63 @@
+<?php
+if ( ! defined( 'ABSPATH' ) ) {
+	exit; // Exit if accessed directly.
+}
+
+?>
+<p>
+	<label for="<?php echo esc_attr( $this->get_field_id( 'title' ) ); ?>"><?php _e( 'Title:', 'business-directory-plugin' ); ?></label>
```
stephywells, line 8: Let's escape all of the translated strings with esc_html_e

includes/views/widget/widget-settings.php
```diff
@@ -0,0 +1,63 @@
...
+</p>
+<p>
+	<label for="<?php echo esc_attr( $this->get_field_id( 'number_of_listings' ) ); ?>"><?php _e( 'Number of listings to display:', 'business-directory-plugin' ); ?></label>
+	<input class="widefat" id="<?php echo esc_attr( $this->get_field_id( 'number_of_listings' ) ); ?>" name="<?php echo esc_attr( $this->get_field_name( 'number_of_listings' ) ); ?>" type="text" value="<?php echo esc_attr( $this->get_field_value( $instance, 'number_of_listings' ) ); ?>" />
+</p>
+<?php $this->_form( $instance ); ?>
+
+<?php if ( in_array( 'images', $this->supports ) ) : ?>
+
+	<?php $style = 'style="' . ( $this->get_field_value( $instance, 'show_images' ) ? '' : 'display: none;' ) . '"'; ?>
```
stephywells, line 19: This makes it tricky to escape the "echo $style" lines later. How about this?
`$style = $this->get_field_value( $instance, 'show_images' ) ? '' : 'display: none';`

Then when it's used:
`style="<?php echo esc_attr( $style ); ?>"`

Or better yet, we should use the "hidden" class.

includes/widgets/class-listings-widget.php
```diff
@@ -4,150 +4,214 @@
...
+				$size_class = $img_size;
+				$attribute  = '';
+				if ( is_array( $size_class ) ) {
+					$width      = $size_class[0] . 'px';
+					$height     = $size_class[1] . 'px';
+					$size_class = implode( 'x', $size_class );
+					$attribute  = "width='$width' height='$height'";
+				}
+				$class      = "attachment-$size_class size-$size_class listing-image";
+				$image_link = '<a href="' . get_permalink( $post->ID ) . '"><img src="' . WPBDP_ASSETS_URL . 'images/default-image-big.gif" class="' . $class . '" ' . $attribute . ' /></a>';
```
stephywells, line 211: Let's use the coming soon image here from the settings.

includes/views/widget/widget-settings.php
```diff
@@ -0,0 +1,63 @@
...
+	<h4><?php _e( 'Thumbnails', 'business-directory-plugin' ); ?></h4>
+	<p>
+		<input id="<?php echo esc_attr( $this->get_field_id( 'show_images' ) ); ?>" class="wpbdp-toggle-images" name="<?php echo esc_attr( $this->get_field_name( 'show_images' ) ); ?>" type="checkbox" value="1" <?php checked( $this->get_field_value( $instance, 'show_images' ), true ); ?> /> <label for="<?php echo esc_attr( $this->get_field_id( 'show_images' ) ); ?>"><?php _e( 'Show thumbnails', 'business-directory-plugin' ); ?></label>
+	</p>
+	<p class="thumbnail-width-config" <?php echo $style; ?> >
```
stephywells, line 24: I'm thinking the width and height settings look like overkill. Let's just do what looks good here with a 33% width on the box with the image, and max 100% width on the image.

includes/views/widget/widget-settings.php
theRealRizeo: Used `hidden` class and simplified the check on this section

includes/widgets/class-listings-widget.php
theRealRizeo: Done, made changes to the helper class to make the function public

theRealRizeo:
Hey @garretlaxton reminder on review of this :)

includes/widgets/class-listings-widget.php
```diff
@@ -4,150 +4,196 @@
...
+
+		return sprintf( $template, $args['html_class'], $html_image, $listing_title );
+	}
+
+	private function render_image( $listing, $args ) {
+		$image_link = '';
+		if ( $args['show_images'] ) {
+			$img_size = 'wpbdp-thumb';
+			if ( $img_id = $listing->get_thumbnail_id() ) {
+				$image_link = '<a href="' . $listing->get_permalink() . '">' . wp_get_attachment_image( $img_id, $img_size, false, array( 'class' => 'listing-image' ) ) . '</a>';
```
stephywells, line 191: Looks like there's a lot of escaping missing here too. Can you please review all the code in this class for escaping?

theRealRizeo: Added escapes

theRealRizeo:
@garretlaxton I have corrected this

assets/css/less/admin.less
```diff
@@ -791,3 +791,13 @@ body.post-type-wpbdp_listing {
 .select2-search__field {
 	width: 100% !important;
 }
+.widget-inside {
+    .widget-fields {
```
stephywells, line 795: Let's use a prefixed class here. A 1px #c3c4c7 border would be nice too.

stephywells (changes requested):

stephywells (approved):
I went ahead and finished this one off. Paul, maybe you can take a look at the changes I made?
https://github.com/Strategy11/business-directory-plugin/pull/39/commits

I didn't realize the image height and width were existing settings. So I added them back in so we don't break things.

theRealRizeo:
Thanks :)

PR #41, opened 2021-10-07
Consistent button appearance

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/2606

- When the "Override WP theme button styling" button is checked, all the BD buttons should use the color in the settings. If not checked, the color comes from the theme.
- This also switches the links under the search form to links instead of input (new listing, manage listings...)
- When the "all listings" and "directory" buttons show, exclude them if already on that page
- The Cancel buttons are now links in the search form and report listing form

I've checked this in Astra, Twenty twenty one, Twenty Twenty. If a theme doesn't style the "button" class, they'll appear as links instead of buttons.


theRealRizeo:
add-listing

stephywells:
Thanks to you both!

Garret, those issues are fixed now.
Paul, I see you created a separate PR, so I'll go ahead and merge this one now.

PR #44, opened 2021-10-11
Request reviews

theRealRizeo:
This is related to https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4765
initial-review-screen

yes-review-screen


theRealRizeo:
@garretlaxton for testing I think you will need 50 or more listings. I edited the code to get the screenshots required :) I think it would be good to have a test script in future to add dummy listings to make testing easier

garretlaxton (approved):
Looks good! I changed the code as well to get this to execute easier.
Screenshot 2021-10-11 at 15-27-55 Directory ‹ Guitar World — WordPress
Screenshot 2021-10-11 at 15-28-05 Directory ‹ Guitar World — WordPress

includes/admin/views/review.php
```diff
@@ -0,0 +1,70 @@
...
+			$( '.wpbdp-review-notice' ).remove();
+		} );
+
+
+		$('.show-wpbdp-feedback').click( function( e ){
+			e.preventDefault();
+			var link = $(this).data('link');
+			var className = '.wpbdp-' + link + '-request';
+			jQuery('.wpbdp-satisfied').hide();
+			jQuery(className).show();
```
stephywells, line 59: Looks like this section went missing:
https://github.com/Strategy11/formidable-forms/blob/master/classes/views/shared/review.php#L36-L42

I'm setting up a new site for this. The URL will be:
https://services.strategy11.com/wp-json/frm/v2/forms/bd-feedback?return=html

includes/admin/views/review.php
```diff
@@ -0,0 +1,70 @@
...
+}
+?>
+<div class="notice notice-info is-dismissible wpbdp-review-notice">
+	<div class="wpbdp-satisfied">
+		<p>
+			<?php echo esc_html( $title ); ?>
+			<br/>
+			<?php esc_html_e( 'Are you enjoying Business Directory Plugin?', 'business-directory-plugin' ); ?>
+		</p>
+		<a href="#" class="wpbdp_reverse_button wpbdp_animate_bg show-wpbdp-feedback wpbdp-button-secondary" data-link="feedback">
```
stephywells, line 13: We're missing the button styling here. Here's what it looks like in Formidable:
Screen Shot 2021-10-11 at 5 43 40 PM
Screen Shot 2021-10-11 at 5 43 30 PM
Screen Shot 2021-10-11 at 5 43 23 PM

stephywells (changes requested):
Thanks for adding this @theRealRizeo ! Looks like a few things got missed.

includes/models/class-reviews.php
```diff
@@ -0,0 +1,162 @@
...
+		include WPBDP_PATH . 'includes/admin/views/review.php';
+	}
+
+
+	/**
+	 * If there are already later requests, don't add it to the inbox again.
+	 *
+	 * @since 4.05.02
+	 */
+	private function has_later_request( $requests, $asked ) {
```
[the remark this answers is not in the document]
theRealRizeo: Yes, I have cleared unused code :)

assets/css/less/admin-notice-review.less
```diff
@@ -0,0 +1,75 @@
+.wpbdp-review-notice {
+    --primary-color: rgb(65, 153, 253);
```
stephywells, line 2: This is an improvement, but the problem is not it doesn't match BD. How about we use the new button styles for BD instead?
https://www.figma.com/file/hey3bHjXZhJpGGt277mHbg/Business_directory?node-id=247%3A598

It would be best if the styles for these buttons go in the admin.less file or something, since they will be used across the admin area.

stephywells (changes requested):
Thanks for your work on this! I think we'll go ahead and get the release out ,and include this in the next one since it's not ready yet.

includes/admin/views/review.php
```diff
@@ -0,0 +1,96 @@
...
+			nonce: '<?php echo esc_html( wp_create_nonce( 'wpbdp_dismiss_review' ) ); ?>'
+		} );
+	}
+	function wpbdpapiGetData( frmcont ) {
+		jQuery.ajax({
+			dataType:'json',
+			url:frmcont.data('url'),
+			success:function(json){
+				var form = json.renderedHtml;
+				form = form.replace(/<script\b[^<]*(community.formidableforms.com\/wp-includes\/js\/jquery\/jquery)[^<]*><\/script>/gi, '' );
```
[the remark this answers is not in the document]
theRealRizeo: @stephywells the class of the form rendered from https://services.strategy11.com/wp-json/frm/v2/forms/bd-feedback?return=html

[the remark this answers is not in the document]
stephywells: Yeah, those classes will stay as is. That's the form source code.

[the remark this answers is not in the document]
theRealRizeo: Screenshot 2021-10-12 182838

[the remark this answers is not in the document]
theRealRizeo: Nevermind, resolved :)
Screenshot 2021-10-12 192422

includes/models/class-reviews.php
```diff
@@ -0,0 +1,149 @@
...
+		// Only show review request if the site has collected enough entries
+		if ( $entries < $count ) {
+			// check the entry count again in a week
+			$this->review_status['time'] = time();
+			update_user_meta( $user->ID, $this->option_name, $this->review_status );
+
+			return;
+		}
+
+		if ( $entries <= 100 ) {
```
truongwp, line 99: I think we can move this `if ... else` to another method. We want to keep the function size smaller. And it would help a lot for readability and future changes.

(Steph recommended this book https://enos.itcollege.ee/~jpoial/oop/naited/Clean%20Code.pdf)

assets/css/less/admin-notice-review.less
```diff
@@ -0,0 +1,112 @@
...
+			display: inline-block;
+
+			position: absolute;
+			background: transparent;
+			border: 1px solid #fff;
+			border-top-color: transparent;
+			border-left-color: transparent;
+			border-bottom-color: @wpbdp-primary-button-hover;
+			border-right-color: @wpbdp-primary-button-hover;
+			border-radius: 50%;
```
truongwp, line 88: Some lines in this block look like duplicated with https://github.com/Strategy11/business-directory-plugin/blob/master/assets/css/less/admin.less#L466

I think there would be a way to reuse them. I am not quite familiar with LESS.

Please add space before `{` as well.

includes/admin/views/review.php
```diff
@@ -0,0 +1,94 @@
...
+		<p><?php esc_html_e( 'Sorry to hear you aren\'t enjoying building with Business Directory Plugin. We would love a chance to improve. Could you take a minute and let us know what we can do better?', 'business-directory-plugin' ); ?></p>
+
+		<div id="wpbdpapi-feedback" class="wpbdpapi-form" data-url="https://services.strategy11.com/wp-json/frm/v2/forms/bd-feedback?return=html&exclude_script=jquery">
+			<span class="wpbdp-wait wpbdp_visible_spinner"></span>
+		</div>
+	</div>
+</div>
+<script type="text/javascript">
+	jQuery( document ).ready( function( $ ) {
+		$(document).on( 'click', '.wpbdp-dismiss-review-notice, .wpbdp-review-notice .notice-dismiss', function( event ) {
```
truongwp, line 46: Many space characters missing in the JS code. Please follow WordPress JS coding standard: https://developer.wordpress.org/coding-standards/wordpress-coding-standards/javascript/

Variables should be declared at the top of each function.

assets/css/less/admin-notice-review.less
theRealRizeo: The less will be cleaned later. There are too many variables that can be reused and the layout does not really follow the less rules. For example, I think global common variables should all be in one file

includes/admin/views/review.php
theRealRizeo: The JS in this case cannot have the variables declared at the top as they are used for DOM checks and used once where applicable

assets/css/less/admin-notice-review.less
truongwp: @theRealRizeo It's ok.

includes/admin/views/review.php
theRealRizeo: I think to prevent js variable conflicts, in this case it would be better not to have them global as we do not know what else will be called or will use the same variable name. 
The modal is loaded in the admin notification, so we would need to define custom js variables, which at this case I do not think are necessary. It would be best to keep them as local variables in this instance

includes/models/class-reviews.php
```diff
@@ -143,7 +144,28 @@ public function dismiss_review() {
 		$review['dismissed'] = $dismissed === 'done' ? true : 'later';
 		$review['asked']     = isset( $review['asked'] ) ? $review['asked'] + 1 : 1;
 
-		update_user_meta( $user_id, $this->option_name, $review );
+		$this->update_user_meta( $user_id, $review );
 		wp_die();
 	}
+
+	/**
+	 * Update user meta
```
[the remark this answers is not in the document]
truongwp: Sorry, maybe it wasn't clear again.

The period should not be added if @param or @result doesn't have description: https://github.com/Strategy11/business-directory-plugin/pull/44/commits/89624e799ead1f071245085ed033876cd20449b0#diff-23206d8c188d0126b7d5f68f872c6d808a09f9138530dd5dff97a55b9f7979e2R164

The description of @param should start with an uppercase, and no `-`.

Please read the inline documentation link I gave.

includes/models/class-reviews.php
```diff
@@ -149,8 +181,10 @@ public function dismiss_review() {
 	/**
 	 * Update user meta.
 	 *
-	 * @param int   $user_id - the user id.
-	 * @param array $review - the review.
+	 * @param int   $user_id. The user id.
```
truongwp, line 180: We should remove the period after `$user_id`. The period after `The user id` is correct.

Please do the same for other @param and @return lines as well.

includes/helpers/class-app.php
```diff
@@ -145,6 +145,49 @@ public static function is_admin_page( $page = 'wpbdp_settings' ) {
...
+		global $pagenow;
+
+		if ( $pagenow !== 'post.php' && $pagenow !== 'post-new.php' && $pagenow !== 'edit.php' ) {
+			return false;
+		}
+
+		$post_type = wpbdp_get_var( array( 'param' => 'post_type' ) );
+
+		if ( empty( $post_type ) ) {
+			$post_id = wpbdp_get_var( array( 'param' => 'post', 'sanitize' => 'absint' ) );
```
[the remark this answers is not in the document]
truongwp: Please align them again.

includes/admin/class-admin.php
```diff
@@ -1068,6 +1072,20 @@ public function check_for_required_pages() {
...
+         */
+        private function maybe_request_review() {
+            WPBDP_Reviews::instance()->review_request();
+        }
+
+        /**
+         * Dismiss review
+         */
+        public function maybe_dismiss_review() {
+            WPBDP_Reviews::instance()->dismiss_review();
```
truongwp, line 1086: We should call `check_ajax_referer()` before this line.

includes/models/class-reviews.php
```diff
@@ -0,0 +1,214 @@
...
+		include WPBDP_PATH . 'includes/admin/views/review.php';
+	}
+
+	/**
+	 * Save the request to hide the review.
+	 *
+	 * @since x.x
+	 */
+	public function dismiss_review() {
+		check_admin_referer( 'wpbdp_dismiss_review' );
```
truongwp, line 140: If `dismiss_review()` could be used in other places, this call could be good. But if it's called by the `maybe_dismiss_review()` method only, we can remove this line because we called `check_ajax_referer()` in `maybe_dismiss_review()` method.

includes/models/class-reviews.php
```diff
@@ -0,0 +1,214 @@
...
+
+		$user_id = get_current_user_id();
+		$review  = $this->get_user_meta( $user_id );
+		if ( empty( $review ) ) {
```
truongwp, line 142: We can move this `if ( empty( $review ) )` block to the below of `if ( isset( $review['dismissed'] ) && 'done' === $review['dismissed'] )` block.

includes/admin/class-admin.php
```diff
@@ -1068,6 +1072,21 @@ public function check_for_required_pages() {
...
+         */
+        private function maybe_request_review() {
+            WPBDP_Reviews::instance()->review_request();
+        }
+
+        /**
+         * Dismiss review
+         */
+        public function maybe_dismiss_review() {
+            check_admin_referer( 'wpbdp_dismiss_review' );
```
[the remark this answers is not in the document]
truongwp: @theRealRizeo I don't see any new commit after you resolve this. Did you forget to push?

assets/css/less/admin.less
```diff
@@ -9,10 +9,12 @@
 	--orange: #F15A24;
 }
 
+@import 'colors.less';
```
stephywells, line 12: I don't think there will be enough in this file to make it worth having. Can the colors go in the existing common.less file instead?

assets/css/less/admin-notice-review.less
```diff
@@ -0,0 +1,112 @@
...
+			font-size: 14px;
+			max-width: 600px;
+			margin-left: auto;
+			margin-right: auto;
+		}
+	}
+	a {
+		margin: 5px;
+	}
+	.wpbdp-button {
```
stephywells, line 23: We'll be using these button styles for the whole admin area, so let's include them in a different file.the wait spinner will also be used admin-wide so we should move that as well. Anything else?

Mentioned here as well:
https://github.com/Strategy11/business-directory-plugin/pull/44#discussion_r727229907

stephywells (changes requested):
Once all the code is done, I'll take another look at the way it works.

assets/css/less/admin-notice-review.less
theRealRizeo: I will move buttons to their own less file, fore easier management to allow use in the admin area

assets/css/less/admin.less
```diff
@@ -9,11 +9,14 @@
...
 @import 'common.less';
 @import 'fonts.less';
+@import 'buttons.less';
```
stephywells, line 15: I think this would be more clear if we call is admin-buttons since we'll only add this styling on the back-end.

assets/css/less/colors.less
```diff
@@ -0,0 +1,7 @@
+@red: #d64226;
+@gray: #d2d2d2;
+@white: #fff;
+@brand: #6D87B9;
+@default-color: #569AF6;
+@wpbdp-primary-button-color: rgb(126,146,188);
```
stephywells, line 6: Can we make this name shorter to make it easier to use? Maybe just primary and secondary?

includes/class-assets.php
```diff
@@ -250,13 +250,17 @@ public function enqueue_css_override() {
         }
     }
 
-	public function enqueue_admin_scripts( $force = false ) {
+    /**
+     * Load resources on admin page
+     *
+     * @param string $hook The current admin page.
+     */
+	public function enqueue_admin_scripts( $hook ) {
```
stephywells, line 258: This is a public function, so we can't remove the $force parameter without digging thoroughly for what issues it may cause.

theRealRizeo: It may need to change, as the button styling was also overriding those in other wp-admin pages. Ideally the assets should only load in BD admin pages. I will look for a workaround for the `force` param

stephywells: Sounds like the classes in the css need to change if are overriding buttons that aren't ours.

PR #45, opened 2021-10-11
Price plan button states

theRealRizeo:
This attempts to fix the pricing buttons to match the styling setting. On check, the text will change to "Selected" . IT also changes the button text to translated inline CSS pseudo
plan-price-button

### DIFF
--- a/includes/controllers/pages/class-submit-listing.php
+++ b/includes/controllers/pages/class-submit-listing.php
@@ -73,7 +73,7 @@ class WPBDP__Views__Submit_Listing extends WPBDP__Authenticated_Listing_View {
             )
         );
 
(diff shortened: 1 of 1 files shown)


includes/class-assets.php
```diff
@@ -77,6 +77,15 @@ public function register_common_scripts() {
         $this->register_select2();
 
 		wp_register_style( 'wpbdp-base-css', WPBDP_ASSETS_URL . 'css/wpbdp.min.css', array(), WPBDP_VERSION );
+
+		$custom_css = "
```
stephywells, line 81: Cool idea Paul! Is it possible to only include this with the plan price template where we use it?

PR #46, opened 2021-10-11
Check license status while checking for updates

stephywells:
This cuts down on external calls to our server. Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4957
I think this will help for now, and we don't need to deprecate the whole file yet. This way, people without the extra plugin will still get updates. It would be good to add a message to let people know to install BD Premium, but that isn't included here.

This code is only used when Business Directory Premium is not installed, or for single licenses that aren't in a bundle.

- Remove the daily check for each individual license
- Check the license status while checking if there are updates
- Switch to option instead of transient for better reliability. On some servers, the transient is always empty

To test:
- with and without Business Directory Premium
- activating/deactivating a plugin individually
- the server tab in debug tools will list any external requests. BD should not be included more than once a day. (testing the page reload a couple times should be fine)

### DIFF
--- a/includes/licensing.php
+++ b/includes/licensing.php
@@ -31,6 +31,7 @@ class WPBDP_Licensing {
 
 		add_action( 'admin_init', array( &$this, 'add_modules_hooks' ) );
         add_filter( 'pre_set_site_transient_update_plugins', array( $this, 'inject_update_info' ) );
+		add_filter( 'plugins_api', array( $this, 'module_update_information' ), 10, 3 );
 
 		$this->register_dismissable();
     }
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
I don't see anymore spam API calls when testing this. This slowed down my localhost, now it's back to normal speed! Tested with/without the premium module and activated & deactivated modules & licenses.
Screenshot 2021-10-11 at 17-19-20 Add Listing – Just another WordPress site

stephywells:
Thanks so much @garretlaxton !

PR #51, opened 2021-10-13
Add document ready check for script

theRealRizeo:
Fixes image edit on the backend admin edit screen


stephywells (changes requested):
Can you please check the code styling on this? Thanks for looking for other spots this may be an issue!

PR #52, opened 2021-10-13
Fix twenty twenty one theme styling

theRealRizeo:
Screenshot 2021-10-13 180927
Screenshot 2021-10-13 180945


assets/css/less/wpbdp.less
```diff
@@ -1,4 +1,5 @@
 @import 'common.less';
+@import 'twenty-twenty-one.less';
```
stephywells, line 2: Maybe we can keep this a bit more generic so we can use it for other themes too?

assets/css/less/twenty-twenty-one.less
```diff
@@ -0,0 +1,17 @@
+.wpbdp-wp-theme-twentytwentyone {
+    .wpbdp-main-box {
+        .submit-btn input[type="submit"] {
+            padding: 12px;
```
stephywells, line 4: 14px 12px looks a bit better for me. This makes the fields and buttons line up.

stephywells (changes requested):

assets/css/less/wpbdp.less
theRealRizeo: It looks well on the other themes, but since twenty twenty one has smaller content area and different styles, it cannot be used on all themes unless we change the css layout of this section

assets/css/less/twenty-twenty-one.less
stephywells: In your screenshot, the find listings button is a little taller than the input. With padding: 14px 12px, it lines up vertically too.

assets/css/less/wpbdp.less
theRealRizeo: File name has been updated :)

assets/css/less/twenty-twenty-one.less
stephywells: @theRealRizeo this is too short now and isn't lining up vertically... Please make sure they line up perfectly since we're overriding the theme styling here.

stephywells (approved):
I didn't look at this again, but if you're sure it lines up, feel free to merge.

PR #55, opened 2021-10-14
Deleted objects error handling

theRealRizeo:
Reported error :

```
An error of type E_ERROR was caused in line 65 of the file /nas/content/live/greennrginstit/wp-content/plugins/business-directory-plugin/includes/admin/payments.php. Error message: Uncaught Error: Call to a member function update() on bool in /nas/content/live/greennrginstit/wp-content/plugins/business-directory-plugin/includes/admin/payments.php:65
Stack trace:
#0 /nas/content/live/greennrginstit/wp-content/plugins/business-directory-plugin/includes/admin/class-admin-controller.php(55): WPBDP__Admin__Payments->payment_update()
```

When a user deleted a payment directly from the database and had this recurring error occur multiple times. This provides better error handling when payments for listings are deleted directly or do not exist


theRealRizeo:
@truongwp thanks for the feedback :) I have set the checks where necessary. The main check is from when the payment is being queried from the database using `WPBDP_Payment::objects()->get`

includes/admin/payments.php
```diff
@@ -132,4 +134,12 @@ function ajax_delete_note() {
         $res->send();
     }
 
+    private function handle_payment_not_found_redirect( $payment ) {
+		if ( ! $payment ) {
+            // Not found.
+            wp_redirect( esc_url_raw( admin_url( 'admin.php?page=wpbdp_admin_payments&wpbdp-view=details' ) ) );
```
stephywells, line 140: Looks like there are spaces here instead of tabs

includes/admin/payments.php
```diff
@@ -132,4 +134,12 @@ function ajax_delete_note() {
         $res->send();
     }
 
+    private function handle_payment_not_found_redirect( $payment ) {
+		if ( !$payment ) {
```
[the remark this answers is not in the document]
theRealRizeo: @stephywells added a comment on that line why the check is not applied there

includes/class-abandoned-payment-notification.php
```diff
@@ -42,6 +42,9 @@ public function send_abandoned_payment_notifications() {
             }
 
             $payment = WPBDP_Payment::objects()->get( $data->id );
+			if ( !$payment ) {
```
stephywells, line 45: The phpcs here is still wrong...

theRealRizeo: I'll need to correct PHPCS formatting for the entire file. As it uses spaces as opposed to tabs

theRealRizeo: Unless I change to spaces from tabs, as most of the classes and files uses spaces as opposed to tabs

stephywells: The tabs here are fine. The problem is "!$payment". This is the same issue we've been talking about here:
https://github.com/Strategy11/business-directory-plugin/pull/55#discussion_r731130874

PR #56, opened 2021-10-18
Feature: Allow shortcodes working in private pages

theRealRizeo:
This allows shortcodes to work in private pages

### DIFF
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -9,9 +9,10 @@ function wpbdp_get_version() {
 
 /**
  * Check pages that have the shortcode and are public.
- * This is used in the main query to ensure that the views for the main pages are loaded
+ * This is used in the main query to ensure that the views for the main pages are loaded.
+ * Allow shortcodes to be used in private pages.
  *
- * @since x.x Allow shortcodes to be used in private pages.
+ * @since x.x
  */
 function _wpbdp_page_lookup_query( $page_id, $count = false ) {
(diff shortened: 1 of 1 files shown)


includes/functions.php
```diff
@@ -7,6 +7,12 @@ function wpbdp_get_version() {
     return WPBDP_VERSION;
 }
 
+/**
+ * Check pages that have the shortcode and are public.
+ * This is used in the main query to ensure that the views for the main pages are loaded
+ *
+ * @since x.x Allow shortcodes to be used in private pages.
```
stephywells, line 14: I'm not sure if our package code will be able to replace the version number if there is extra text in the line. Can you move the extra text up?

includes/functions.php
```diff
@@ -22,9 +28,9 @@ function _wpbdp_page_lookup_query( $page_id, $count = false ) {
     }
 
     if ( $count ) {
-        $query = "SELECT COUNT(*) FROM {$wpdb->posts} WHERE post_type = 'page' AND post_status = 'publish' AND ( 1=0";
+        $query = "SELECT COUNT(*) FROM {$wpdb->posts} WHERE post_type = 'page' AND post_status IN ( 'publish', 'private' ) AND ( 1=0";
```
stephywells, line 31: I don't think we need the post status check here at all. It should show if previewing a draft too.

PR #58, opened 2021-10-20
Listing image performance improvements

theRealRizeo:
Reported https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4979

- Added second argument to the function `_wpbdp_resize_image_if_needed` called `$args` that allows of passing of width, height and crop arguments as an array to prevent querying the options when function is used in a loop
- Added a filter `wpbdp_resize_image_if_needed` that accepts the arguments `$resize (bool)`, `$id`, `$args` . The filter can be used to skip resizing of a specific or all images if the site owner does not want the images to be resized. By default it returns `true` to allow resizing

### DIFF
--- a/includes/helpers/class-listing-display-helper.php
+++ b/includes/helpers/class-listing-display-helper.php
@@ -275,7 +275,7 @@ class WPBDP_Listing_Display_Helper {
         $crop           = wpbdp_get_option( 'thumbnail-crop' );
         foreach ( $listing_images as $img_id ) {
             // Correct size of thumbnail if needed.
-            _wpbdp_resize_image_if_needed( 
+            _wpbdp_resize_image_if_needed(
                 $img_id,
                 array(
                     'width'  => $def_width,
--- a/includes/functions.php
+++ b/includes/functions.php
@@ -477,14 +477,14 @@ function _wpbdp_should_image_be_resized( $id, $args = array() ) {
 		return false;
 	}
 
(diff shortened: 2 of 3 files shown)


stephywells (commented):
@theRealRizeo it looks like the issue is that large images are being used for thumbnails in some cases. Can you tell me how this addresses the issue?

truongwp (commented):
I didn't check this issue. I will almost focus on the coding standard and improvements.

Can we move these lines to a separate function?
https://github.com/Strategy11/business-directory-plugin/pull/58/files#diff-1c08b27bcf43583c75cbc442bc401638be3427586fe544f3389e22c9979fa1afR441-R475

Look like those lines are only used to check if we should resize the image.

includes/helpers/class-listing-display-helper.php
```diff
@@ -270,10 +270,16 @@ private static function images_vars( $listing_id, $display ) {
...
         // Other images.
         $listing_images = $listing->get_images( 'ids', true );
-
+        $def_width      = wpbdp_get_option( 'thumbnail-width' );
+        $def_height     = wpbdp_get_option( 'thumbnail-height' );
+        $crop           = wpbdp_get_option( 'thumbnail-crop' );
         foreach ( $listing_images as $img_id ) {
             // Correct size of thumbnail if needed.
-            _wpbdp_resize_image_if_needed( $img_id );
+            _wpbdp_resize_image_if_needed( $img_id, array(
```
truongwp, line 278: Two params `$img_id` and `array` should be on their own line:

```
function(
    $img_id,
    array(
        .....
    )
)
```

PR #59, opened 2021-10-20
Add class to renew listing

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4980

### DIFF
--- a/includes/controllers/pages/class-submit-listing.php
+++ b/includes/controllers/pages/class-submit-listing.php
@@ -32,7 +32,13 @@ class WPBDP__Views__Submit_Listing extends WPBDP__Authenticated_Listing_View {
         return __( 'Add Listing', 'business-directory-plugin' );
     }
 
-    public function enqueue_resources() {
+    /**
+     * Load custom resources used for this view only.
+     * This is called in the parent class `enqueue_resources` function.
+     *
+     * @since x.x
+     */
+    public function enqueue_custom_resources() {
         wp_enqueue_style( 'dashicons' );
 
         wp_enqueue_script(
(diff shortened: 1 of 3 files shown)


theRealRizeo:
Yes, I have made some changes to allow code re-use for the inline css

includes/helpers/class-authenticated-listing-view.php
```diff
@@ -2,6 +2,35 @@
...
+
+        $this->enqueue_custom_resources();
+    }
+
+    /**
+     * Load custom resources in classes that extend this class
+     *
+     * @since x.x
+     */
+    public function enqueue_custom_resources() {
```
stephywells, line 30: Empty function here

includes/controllers/pages/class-submit-listing.php
```diff
@@ -32,7 +32,13 @@ public function get_title() {
...
     }
 
-    public function enqueue_resources() {
+    /**
+     * Load custom resources used for this view only.
+     * This is called in the parent class `enqueue_resources` function.
+     *
+     * @since x.x
+     */
+    public function enqueue_custom_resources() {
```
stephywells, line 41: This function name was changed. Since it's a public function, we can't just remove it unless we deprecate the previous function.

includes/helpers/class-authenticated-listing-view.php
theRealRizeo: Explained below

stephywells: There should be a comment here for the empty function. Also, be sure to use tabs instead of spaces.

theRealRizeo: Changes uploaded. Fixed space indentation in the rest of the file as well

garretlaxton (approved):
Apologies, I was trying to select an invalid plan... duh

includes/helpers/class-authenticated-listing-view.php
theRealRizeo: For empty I have seen above is only if there are params like https://github.com/WordPress/WordPress/blob/4152fbfdb2e41a5de19b664a666e6620f13499b4/wp-admin/includes/class-wp-list-table.php#L1403 but most common is having the comment in the empty function https://wordpress.stackexchange.com/questions/145569/actions-functions-and-conditionals

stephywells (approved):
Looks good now! Thanks Paul!

PR #65, opened 2021-10-31
Fee plan active and inactive states

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4989

My understanding is:

- Payments on, default plan should be disabled
- Payments off, default plan should be enabled

The default plan cannot be disabled or enabled from the code, but it obeys the above states. So from the admin page, the disabled/enabled buttons are removed.
If payments are  turned off, paid plans cannot be disabled or enabled.

UPDATE:

- Toggle for payments has been removed
- Default enabled paid plans will be disabled if payments are off
- Default enabled free plans will be disabled if payments are on
- Default plan can be enabled or disabled
- Returned the filter for "enabled" and "disabled" in the fee plans. Makes it easier to sort the fee plans

Tests Updates:

- Tests create free and premium plans. Some enabled and some disabled
- Tests run with payments enabled by default
- Tests run with payments disabled by default
- New base classes for plan tests and listing tests


garretlaxton (approved):
This works as described. When the paid mode is on, the default-free plan is disabled and when paid mode is off, the default plan is enabled while all other plans are disabled. I like this logic.

I was curious how renewals would work with disabled plans and you can still renew a listing with a disabled plan. If you have a listing in a plan that gets disabled and you go to renew, you can just click continue and it will renew with the previous fee plan disabled or not. I'm guessing we would need some type of check during renewal to see if the plan is disabled or not. We can probably log this separately but figured it's worth mentioning.

theRealRizeo:
@garretlaxton I think its why the option to enable/disable payments will be removed in future. It would make sense just to have plans with prices and those without, where a user enables or disables  them

garretlaxton:
@theRealRizeo I agree, that would be a lot cleaner.

stephywells:
I'm thinking we just keep adding bandaids to this. How about we just add a migration in here so we can do it right? Here's what it would mean:
- If a plan is already marked as disabled, it can be ignored in the migration
- If payments are turned on, the default plan should be switched to disabled and remove the "free" tag
- If payments are turned off, any paid plans should be switched to disabled
- The "free" tag for plans should be removed from the code
- The setting to turn on payments should be removed. This can be replaced with a function that checks for active plans with a price
- Let's include unit tests for the migration

theRealRizeo:
@stephywells changes done except removing of the `free` tag. The `free` tag is used to identify the default plan that is assigned to a listing during import or creation (admin side) when there is no plan found. Removing it would have bigger side effects and may alter how the plugin saves plans. A default plan needs to remain like a default category

Regarding the test case, I have added the test but excluded it as there are mysql connection issues I am working on in the `test-cases` branch with the options and other WPUnit configurations

garretlaxton:
This functions very well. One little thing I noticed is when you enable a fee plan it shows `Fee Disabled` rather than `Fee Enabled`.
Screenshot 2021-11-08 at 17-31-31 Fee Plans ‹ BD QA — WordPress

theRealRizeo:
Thanks @garretlaxton , this has been corrected now :)

includes/admin/upgrades/migrations/migration-18_5.php
```diff
@@ -0,0 +1,41 @@
...
+	 * We scan all enabled plans first for this migration.
+	 * If payments was enabled, we disable the free plans.
+	 * If payments was disabled, we disabled the paid plans.
+	 *
+	 * @since x.x
+	 */
+	public function migrate() {
+		global $wpdb;
+		$payments_on = wpbdp_get_option( 'payments-on' );
+		$sql         = "SELECT `id`, `amount` FROM {$wpdb->prefix}wpbdp_plans WHERE `enabled` = %d";
```
stephywells, line 22: I think we need to reverse this to enabled != 0. Plans may get missed this way.

Also, let's be consistent and not use the ``

includes/admin/upgrades/migrations/migration-18_5.php
```diff
@@ -0,0 +1,41 @@
...
+				$to_disable[] = $plan->id;
+			} elseif( ! $payments_on && $plan->amount > 0.0 ) {
+				$to_disable[] = $plan->id;
+			}
+		}
+
+		if ( ! empty( $to_disable ) ) {
+			$sql = "UPDATE {$wpdb->prefix}wpbdp_plans SET `enabled` = 0 WHERE `id` IN(" . implode( ', ', array_fill( 0, count( $to_disable ), '%d' ) ) . ")";
+			// Call $wpdb->prepare passing the values of the array as separate arguments.
+			$query = call_user_func_array( array( $wpdb, 'prepare' ), array_merge( array( $sql ), $to_disable ) );
```
stephywells, line 37: Can we use $wpdb->update here?

includes/admin/upgrades/migrations/migration-18_5.php
```diff
@@ -0,0 +1,41 @@
...
+	 * If payments was disabled, we disabled the paid plans.
+	 *
+	 * @since x.x
+	 */
+	public function migrate() {
+		global $wpdb;
+		$payments_on = wpbdp_get_option( 'payments-on' );
+		$sql         = "SELECT `id`, `amount` FROM {$wpdb->prefix}wpbdp_plans WHERE `enabled` = %d";
+		$plans       = $wpdb->get_results( $wpdb->prepare( $sql, 1 ) );
+		$to_disable  = array();
```
stephywells, line 24: Can you add a return here if there are no plans?

includes/class-payment-gateways.php
```diff
@@ -115,10 +114,6 @@ public function _admin_warnings() {
             return;
         }
 
-        if ( ! wpbdp_get_option( 'payments-on' ) ) {
```
stephywells, line 118: If there aren't any plans with fees, we should nag about adding payment gateways. Let's keep this part.

Can you please go through and test every change on both a new and existing site?

stephywells (changes requested):

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -25,16 +25,39 @@ public function __construct() {
...
+		if ( 'all' === $this->get_current_view() ) {
+            echo str_replace(
+                '<a>',
+                '<a href="' . esc_url( admin_url( 'admin.php?page=wpbdp-admin-fees&wpbdp-view=add-fee' ) ) . '">',
+                _x( 'There are no fees right now. You can <a>create one</a>, if you want.', 'fees admin', 'business-directory-plugin' )
+            );
+            return;
+        }
+
+		$current_view = $this->get_current_view();
```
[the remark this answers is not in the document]
theRealRizeo: Switcher tabs are necessary UX wise. Without it its hard for a user to know what is enabled and what is disabled. There needs to be a way to know states easily for the user. When I was testing I found it hard to know what is enabled and/or disabled

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -1153,23 +1153,13 @@ private static function settings_payment() {
...
-            )
-        );
         wpbdp_register_setting(
             array(
                 'id'           => 'payments-test-mode',
                 'type'         => 'checkbox',
                 'name'         => _x( 'Put payment gateways in test mode?', 'settings', 'business-directory-plugin' ),
                 'default'      => true,
                 'group'        => 'payment/main',
-                'requirements' => array( 'payments-on' ),
```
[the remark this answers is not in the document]
theRealRizeo: It will show a blank page as there is no option to toggle payments. To show conditionally the entire payments tab will need to be hidden. Alternatively we can just leave the payments details open, like in Woo, and give the user the option. This will also by default, show new users that there are payment options available to configure, instead of hiding options from the user in the beginning

includes/admin/upgrades/migrations/migration-18_5.php
theRealRizeo: `$wpdb->update` here would be too slow here for bulk updates and can take more time as opposed to a direct SQL query

includes/admin/helpers/tables/class-fees-table.php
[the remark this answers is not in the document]
stephywells: We can ask Yaro to design a background color or something on the disabled plans. I think having navigation just adds complication. We can't really compare to WooCommerce here, since products could get into the thousands. In this case, the number of plans will stay very small.

Here's a good example I found doing it this way (0:59):
https://bootstrap.brilliantdirectories.com/support/solutions/articles/12000065945-guide-4-how-to-add-membership-plans-in-admin-area

includes/admin/settings/class-settings-bootstrap.php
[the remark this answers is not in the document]
stephywells: Yep, you're right. If there isn't anything else on the page, there's no reason to hide it all.

garretlaxton (approved):
Looks good now!

includes/admin/upgrades/migrations/migration-18_5.php
stephywells: Looks like there's not a way to use "IN" in the update query. Marking this a resolved.

includes/class-payment-gateways.php
theRealRizeo: Tested on new and existing site. Removed notice that asked users to enable or disable payments.

- When a plan with a price is enabled and there are no active gateways, this is shown
Screenshot 2021-11-10 164043
- When a gateway is  enabled, normal checkout flow goes.
- Only enabled plans are shown on checkout

theRealRizeo: Doing more tests on the checkout process

theRealRizeo:
Current bug I am fixing involves a user selecting a paid plan, then going back and selecting a free plan. It should skip payment at this point, but not currently doing that. I am seeing this as a scenario where an admin would like to offer a free plan with limited features and a paid plan with more features

theRealRizeo:
@garretlaxton I've done some changes and tests that need a fresh pair of eyes please :) Please test the following scenarios

- Only free plans enabled (should not ask for payment)
- Only paid plans enabled (should ask for payment)
- Mixture of both free and paid plans (depending on what is selected, it should go through)
- Mixture of both free and paid plans. Select a paid plan, proceed to checkout, press back and change the plan to a free one. Then proceed to complete. You should not see a payment form when the plan is changed from paid to free.
- As above, please try first a free plan, then switch to paid. Checkout form should appear for paid

Not sure if this scenario would come out, but I assume a site owner should be able to offer a free plan with limited features, then a paid option. Please advise if this logic makes sense

garretlaxton (approved):
- Only free plans enabled (should not ask for payment) ✔️
- Only paid plans enabled (should ask for payment) ✔️
- Mixture of both free and paid plans ✔️
- Mixture of both free and paid plans. Paid, then switch back to free. ✔️
- - Mixture of both free and paid plans. Free, then switch back to paid. ✔️

All in all, I am only presented with the checkout screen when a paid plan is selected. Looks great! And yes, your logic makes perfect sense.

theRealRizeo:
Thanks @garretlaxton :)

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -25,12 +25,11 @@ public function __construct() {
...
     public function no_items() {
-		printf(
-			/* translators: %1$s: open link html, %2$s close link */
-			esc_html__( 'There are no fees right now. %1$sCreate one%2$s.', 'business-directory-plugin' ),
+		echo str_replace(
+			'<a>',
 			'<a href="' . esc_url( admin_url( 'admin.php?page=wpbdp-admin-fees&wpbdp-view=add-fee' ) ) . '">',
-			'</a>'
-        );
+			_x( 'There are no fees right now. You can <a>create one</a>, if you want.', 'fees admin', 'business-directory-plugin' )
```
stephywells, line 31: This line went backwards... It was using printf.

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -47,8 +46,9 @@ public function get_views() {
 		$all = absint( $wpdb->get_var( "SELECT COUNT(*) FROM {$wpdb->prefix}wpbdp_plans" ) );
 
         $views['all'] = sprintf(
-            '<a href="%s" class="current">%s</a> <span class="count">(%s)</span></a>',
+            '<a href="%s" class="%s">%s</a> <span class="count">(%s)</span></a>',
```
stephywells, line 49: There are still changes here that didn't get reverted.

includes/admin/upgrades/migrations/migration-18_5.php
```diff
@@ -0,0 +1,47 @@
...
+		$sql         = "SELECT p.id, p.amount FROM {$wpdb->prefix}wpbdp_plans p WHERE p.enabled != 0";
+		$plans       = $wpdb->get_results( $sql );
+		$to_disable  = array();
+
+		if ( ! $plans ) {
+			return;
+		}
+
+		foreach ( $plans as $plan ) {
+			if ( $payments_on && $plan->amount <= 0.0 ) {
```
stephywells, line 31: This logic will change the plans that currently show on the page. When payments are on, the default free plan should be disabled. The other free plans should be left alone.

Please refer back to the migration details listed here:
https://github.com/Strategy11/business-directory-plugin/pull/65#issuecomment-959809227

includes/models/class-listing.php
```diff
@@ -670,6 +670,20 @@ public function get_fee_plan() {
...
+	 * Get the fee plan id.
+	 *
+	 * @since x.x
+	 *
+	 * @return int|bool The plan id or false.
+	 */
+	public function get_fee_plan_id() {
+		global $wpdb;
+
+		$sql = $wpdb->prepare( "SELECT fee_id FROM {$wpdb->prefix}wpbdp_listings WHERE listing_id = %d LIMIT 1", $this->id );
```
stephywells, line 683: This should go through our DB functions with a caching layer.

tests/wpunit/Migrations/FeePlanMigrationTest.php
```diff
@@ -0,0 +1,67 @@
...
+	protected $tester;
+
+	protected function _before() {
+		$this->save_fee();
+	}
+
+	protected function _after() {
+		global $wpdb;
+
+		// Remove any existing payment in database
```
stephywells, line 21: This comment doesn't make sense to me.

tests/wpunit/Migrations/FeePlanMigrationTest.php
```diff
@@ -0,0 +1,67 @@
...
+		$disabled    = absint( $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM {$wpdb->prefix}wpbdp_plans WHERE enabled = %d", 0 ) ) );
+		$enabled     = $all - $disabled;
+
+		Debug::debug( 'Total disabled ' . $disabled );
+		Debug::debug( 'Total enabled ' . $enabled );
+
+		$migrator    = new WPBDP__Migrations__18_5();
+		$migrator->migrate();
+		$after_disabled    = absint( $wpdb->get_var( $wpdb->prepare( "SELECT COUNT(*) FROM {$wpdb->prefix}wpbdp_plans WHERE enabled = %d", 0 ) ) );
+		$after_enabled     = $all - $after_disabled;
```
stephywells, line 41: I don't see any checks to compare actual and expected results here.

stephywells (changes requested):
I'm seeing lots of scary stuff in here that looks like it will change the available plans... This is critical to get right.

So I think we need another branch that adds a unit test that checks the plans that are included in the submit listing form. We should write this for the current code. The unit test should do this:
- Create 2 new free plans
- Create 2 new paid plans
- Get a count of plans that show in the new listing form with payments turned off and make sure it equals 3
- With payments on, get a count and make sure it equals 4

Also, it is critical that the unit tests are running right, or we lose 100% of the benefit. Once that is done, we run those test against this branch too.

tests/wpunit/Migrations/FeePlanMigrationTest.php
theRealRizeo: Test cases on hold until I get a DB function working with the `$wpdb` functions. Its skipped for now

includes/admin/upgrades/migrations/migration-18_5.php
theRealRizeo: The current logic is :

- Payments on : all free plans are disabled
- Payments off : all paid plans disabled and free plans enabled

The code here takes care of that as we do not want a situation after migration to enable free plans on a site that had them disabled due to condition 1

theRealRizeo: I'm still running tests on this logic with a fresh install with default settings and the updated settings applied to it

tests/wpunit/Migrations/FeePlanMigrationTest.php
theRealRizeo: The logic on this is still being worked on and the test case on this is skipped based on other issues

theRealRizeo:
@stephywells noted. I am still conducting tests on this to get it right, I was collecting feedback mainly on the logic of the plans

theRealRizeo:
Closing this as its not ready to be a PR just  yet

includes/admin/upgrades/migrations/migration-18_5.php
stephywells: But that's not the way it currently works. Can you please set up the case I mentioned for unit tests and see how it works in the current version? It's impossible not to break it if you're not super familiar with what should show.

theRealRizeo: Yes, I am currently running various tests cases. This will be adjusted after I finish various tests with the main code and migrations triggered manually

tests/wpunit/Migrations/FeePlanMigrationTest.php
theRealRizeo: Hey @stephywells I have been doing some research on the automated code tests. Generally the error I am facing is this https://githubmemory.com/repo/lucatume/wp-browser/issues/497 . From the CodeCeption Lucatime dev, the resolution is https://wpbrowser.wptestkit.dev/faq#can-i-run-all-my-tests-with-one-command . I have modified the codeception script to run each wpunit file separately for the tests. Running in parallel causes the database connection to be lost as it creates and drops WordPress data on each  test

includes/models/class-listing.php
theRealRizeo: Code removed as its not being used

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -1153,23 +1153,13 @@ private static function settings_payment() {
             )
         );
 
-        wpbdp_register_setting(
-            array(
-                'id'      => 'payments-on',
```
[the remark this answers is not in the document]
theRealRizeo: The setting is not being used. There is a function that checks if the gateways are enabled. For now this setting would only be used in migration

includes/admin/upgrades/migrations/migration-18_5.php
theRealRizeo: Working on the branch now for the additional tests

includes/installer.php
```diff
@@ -36,15 +36,15 @@ public function install() {
         if ( false === get_option( 'wpbdp-db-migrations', false ) )
             update_option( 'wpbdp-db-migrations', array(), false );
 
-        if ( self::DB_VERSION == $this->installed_version )
+		if ( version_compare( self::DB_VERSION, $this->installed_version, '=' ) )
```
theRealRizeo, line 39: @stephywells please let me know if this is okay to compare the DB versions or if I should revert

stephywells: That looks fine to me. Thanks!

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -72,24 +72,21 @@ public function prepare_items() {
...
     public function single_row( $item ) {
-        $free_mode = ( ! wpbdp_payments_possible() );
         $classes   = '';
-
-        if ( $free_mode && $item->amount > 0.0 ) {
-            $classes .= 'disabled-fee';
-        } elseif ( 'free' === $item->tag ) {
-            $classes .= 'free-fee';
-        }
+		if ( $item->is_enabled_premium() ) {
```
stephywells, line 85: This class is getting added to all my plans, not just the disabled ones.

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -72,24 +72,21 @@ public function prepare_items() {
...
+			$classes .= 'disabled-fee';
+		} elseif ( 'free' === $item->tag ) {
+			$classes .= 'free-fee';
```
stephywells, line 88: I don't think we need this class. Maybe you can look more into the js that uses this?
$('.wpbdp-admin-page-fees .wp-list-table.fees tbody').sortable({

Any idea when this table stopped being sortable?

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -1130,6 +1130,16 @@ private static function settings_appearance() {
     }
 
     private static function settings_payment() {
+		wpbdp_register_setting(
+			array(
+				'id'      => 'payments-on',
+				'type'    => 'hidden',
+				'class'   => 'hidden',
+				'default' => false,
```
stephywells, line 1138: Let's switch the default on this to true. I think that would be a better fallback in case add-ons are using this check.

includes/helpers/functions/general.php
```diff
@@ -262,10 +262,6 @@ function wpbdp_get_formfield( $id ) {
 /* Fees/Payment API */
 
 function wpbdp_payments_possible() {
-    if ( ! wpbdp_get_option( 'payments-on' ) ) {
```
stephywells, line 265: Instead of only checking for gateways, we need to replace this line with a check for active plans with a price.

Here's what should happen:
Before: The setting was on => payments are on
After: There is an enabled plan with a price => payments are on

includes/models/class-fee-plan.php
```diff
@@ -269,6 +269,21 @@ public function calculate_expiration_time( $base_time = null ) {
...
+	/**
+	 * Checks if the current plan is premium and enabled.
+	 * This checks if payments are turned on and the price is set.
+	 *
+	 * @since x.x
+	 *
+	 * @return bool
+	 */
+	public function is_enabled_premium() {
+		if ( ( $this->amount > 0.0 ) && ! wpbdp_payments_possible() ) {
```
stephywells, line 281: This looks like it will end up creating a loop since we need to check plans in order to now if payments are on.

It looks like this function should be removed completely, unless it is used to check if the plan has a price:
```
public function is_paid_plan() {
return $this->amount > 0.0;
}
```

includes/class-payment-gateways.php
stephywells: It looks like this still return needs to be added back in along with the gateway check below it.

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -174,6 +171,11 @@ public function column_label( $fee ) {
                 ),
                 esc_html__( 'Delete', 'business-directory-plugin' )
             );
+
+			if ( ! $fee->is_enabled_premium() ) {
```
stephywells, line 175: I'm not really understanding why this is here, after already checking if these should be added. It looks like this should be removed, and don't unset any actions.

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -283,19 +285,13 @@ public function column_categories( $fee ) {
...
 		$tags = array();
 
-		$is_default_free_plan = 'free' === $fee->tag;
-		$is_paid_plan         = 0.0 !== $fee->amount;
-		$payments_on          = wpbdp_payments_possible();
-
-		if ( ! $fee->enabled ) {
-			$tags[] = esc_html__( 'Disabled', 'business-directory-plugin' );
-		} elseif ( ( ! $payments_on && $is_paid_plan ) || ( $payments_on && $is_default_free_plan ) ) {
+		if ( ! $fee->enabled || ! $fee->is_enabled_premium() ) {
```
stephywells, line 288: It looks like this should just be:
`if ( ! $fee->enabled ) {`

includes/controllers/pages/class-submit-listing.php
```diff
@@ -136,17 +136,27 @@ public function dispatch() {
...
-                return $this->_redirect( $possible_payment->get_checkout_url() );
-            }
+			$plan_id = absint( wpbdp_get_var( array( 'param' => 'listing_plan', 'default' => 0 ), 'post' ) );
+			$plan    = wpbdp_get_fee_plan( $plan_id );
+            if ( $plan && $plan->enabled ) {
+				$plan    = $this->get_plan_for_listing();
+				$updated = $this->maybe_update_listing_plan( $plan );
+				if ( $updated ) {
+					// New listing plan changed. We need to delete other payment records.
+					$this->listing->delete_payment_history();
```
stephywells, line 146: Let's never delete payment history for successful payments. The database structure for this is pretty rough. I think the problem here is that the plan id should be saved in the payments table.

stephywells (changes requested):
This is looking much better!

includes/admin/helpers/tables/class-fees-table.php
theRealRizeo: Seems the drag handle is missing, will try locate it

theRealRizeo: The column `order` was missing. Checked all history and it wasn't there. Added it

includes/models/class-fee-plan.php
theRealRizeo: Agreed :)

includes/admin/helpers/tables/class-fees-table.php
theRealRizeo: I've removed this check, it made sense when payments options was present

includes/controllers/pages/class-submit-listing.php
theRealRizeo: Changed label to `in progress` Need to write some tests on the submission part with the plans

theRealRizeo:
@garretlaxton made some code changes, please re-test :)

garretlaxton:
The total price seems to be inconsistent when viewing the plans. The top should be $30, second should be $3, third is correct, and fourth should be $20.
Screenshot 2021-12-15 at 15-59-14 Fee Plans ‹ BD QA — WordPress

theRealRizeo:
@garretlaxton that is the total revenue for the specific fee. It checks the payment table for completed payments only for the specific plan

garretlaxton:
Thanks for that, makes perfect sense. 

One concern with this is if an admin creates a listing they don't have to pay, the price will get added to the total still which can give false results. Could we exclude listings that are `Admin Posted`?

stephywells:
@theRealRizeo where are we on this PR?

theRealRizeo:
@stephywells I will to re-review this with the changes in main, then make the applicable changes

theRealRizeo:
@garretlaxton done some changes to exclude the admin totals. Please test

garretlaxton:
Admin posting listings look good now.

Another thing I found while using the discount codes module. When you create a listing with a 100% off discount code, the total will appear as `Free`.
Screenshot 2022-01-11 at 17-27-08 Fee Plans ‹ BD QA — WordPress

theRealRizeo:
@garretlaxton moving the fee plan table changes to this issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5015

stephywells:
@theRealRizeo I'm not sure who this PR is waiting on right now... Is the In progress tag correct? It would be great if you can keep the tags up to date so we know where in the process the PR is, and who it's currently waiting on. Do we need another tag for QA complete?

theRealRizeo:
Updating the tags, was looking into it yesterday

theRealRizeo:
@garretlaxton I've set this ready for QA

garretlaxton (approved):
Looks good here. The discount issue was fixed in a separate PR.

includes/admin/helpers/tables/class-fees-table.php
```diff
@@ -82,14 +83,12 @@ public function prepare_items() {
...
     public function single_row( $item ) {
-        $free_mode = ( ! wpbdp_payments_possible() );
-        $classes   = '';
-
-        if ( $free_mode && $item->amount > 0.0 ) {
-            $classes .= 'disabled-fee';
-        } elseif ( 'free' === $item->tag ) {
-            $classes .= 'free-fee';
-        }
+        $classes = 'fee';
```
stephywells, line 86: Please check tabs

includes/controllers/pages/class-submit-listing.php
```diff
@@ -136,17 +136,23 @@ public function dispatch() {
...
-					'status'       => 'pending',
-                )
-            )->get();
-
-            if ( $possible_payment ) {
-                return $this->_redirect( $possible_payment->get_checkout_url() );
-            }
+			$plan_id = absint( wpbdp_get_var( array( 'param' => 'listing_plan', 'default' => 0 ), 'post' ) );
+			$plan    = wpbdp_get_fee_plan( $plan_id );
+            if ( $plan && $plan->enabled ) {
```
stephywells, line 141: Please check tabs

includes/controllers/pages/class-submit-listing.php
```diff
@@ -136,17 +136,23 @@ public function dispatch() {
...
+				$plan = $this->get_plan_for_listing();
```
stephywells, line 142: It looks like this line should be removed? This doesn't make sense to me.

includes/controllers/pages/class-submit-listing.php
```diff
@@ -1475,6 +1480,25 @@ private function get_plan_for_listing() {
 		return $listing->get_fee_plan();
 	}
 
+	/**
+	 * Change plan if not the same as for listing.
+	 * Update the plan in the payment.
+	 *
+	 * @param object $plan The plan.
```
stephywells, line 1487: Which plan? Maybe we can rename this to $new_plan?

stephywells (changes requested):
It looks like there are still a few comments from my last review that got missed.

includes/controllers/pages/class-submit-listing.php
theRealRizeo: Agreed

includes/models/class-listing.php
```diff
@@ -812,8 +812,12 @@ public function generate_or_retrieve_payment() {
 
         $existing_payment = WPBDP_Payment::objects()->filter( array( 'listing_id' => $this->id, 'payment_type' => 'initial' ) )->get();
 
-        if ( $existing_payment )
-            return $existing_payment;
+        if ( $existing_payment ) {
+            $key = array_search( $plan->fee_id, array_column( $existing_payment->payment_items, 'fee_id' ), true );
```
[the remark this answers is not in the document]
theRealRizeo: Comment added

includes/class-payment-gateways.php
```diff
@@ -138,12 +138,6 @@ public function _admin_warnings() {
             }
         }
 
-        if ( ! $at_least_one_gateway ) {
```
stephywells, line 141: Was there a reason to remove this message?

includes/helpers/functions/general.php
```diff
@@ -991,6 +995,28 @@ function wpbdp_get_fee_plan( $plan_id ) {
...
 
+/**
+ * Check if there are enabled premium fee plans.
+ * This does a count for all enabled fee plans with a price greater than 0.
+ *
+ * @since x.x
+ *
+ * @return bool
+ */
+function wpbdp_has_premium_fees() {
```
stephywells, line 1006: Is there a place we could put this inside of a class?

stephywells (changes requested):

includes/class-payment-gateways.php
theRealRizeo: With the option to turn payment on/off, the message didn't make sense. Looking at it now, it will need some text change. I will add some

includes/helpers/functions/general.php
theRealRizeo: Moved this to `WPBDP_Fees_API` class, now called as `WPBDP_Fees_API::has_premium_fees()`

includes/class-payment-gateways.php
theRealRizeo: Removed the message from the Fee Plan section. Not sure if we should keep it or change the message. It feels redundant showing the same message in the fee plans page.

stephywells: I made some adjustments here to show each of these messages when it make more sense.

stephywells (commented):
I'm trying to test this, but my checkout page won't load when I try to create a paid listing. Is this working for anyone else? @theRealRizeo @garretlaxton 

It's possible I did something to mess it up.

theRealRizeo:
@stephywells Did a test with paid and free listing, process completed. 
If using Auth.net, it will require valid credentials to load the form

stephywells (approved):
This looks all ready to go now!

PR #67, opened 2021-11-03
Fix abandoned url

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4992

Abandoned url was going to renew url, which should actually just go to the payment url with the same payment id

### DIFF
--- a/includes/controllers/pages/class-manage-listings.php
+++ b/includes/controllers/pages/class-manage-listings.php
@@ -82,6 +82,11 @@ class WPBDP__Views__Manage_Listings extends WPBDP__View {
 
     }
 
+    /**
+     * Show different payment url for a pending listing
+     *
+     * @since x.x
+     */
     public function maybe_add_renew_button( $buttons, $listing_id ) {
         $listing = wpbdp_get_listing( $listing_id );
         $listing_status  = $listing->get_status();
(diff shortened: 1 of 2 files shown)


stephywells:
@theRealRizeo do you know what status an expired listing has? Does it switch to pending_payment too?

theRealRizeo:
@stephywells expired listings have the status `expired`  They are not switched to `pending_payment` from what I have seen

garretlaxton (approved):
Works great, I can now renew abandoned listings.

My question now is, should we change the button text from `Renew Listing` to something like `Pay Now` for abandoned listings?

theRealRizeo:
@garretlaxton this has been added :)

includes/controllers/pages/class-manage-listings.php
```diff
@@ -90,11 +95,12 @@ public function maybe_add_renew_button( $buttons, $listing_id ) {
             return $buttons;
         }
 
+		$is_pending_payment = ( 'pending_payment' === $listing_status );
         $buttons = sprintf(
             '<a class="wpbdp-button button renew-listing" href="%s" %s >%s</a>',
-            $listing->get_renewal_url(),
+            $is_pending_payment ? $listing->get_payment_url() : $listing->get_renewal_url(),
```
stephywells, line 101: This is missing an esc_url

includes/controllers/pages/class-manage-listings.php
```diff
@@ -82,6 +82,11 @@ public function maybe_remove_listing_buttons( $res, $listing_id, $user_id ) {
 
     }
 
+    /**
+     * Show different payment url for a pending listing
+     *
+     * @since x.x
```
stephywells, line 88: Maybe we should remove this since because it won't be accurate?

stephywells (changes requested):
Looking good!

includes/controllers/pages/class-manage-listings.php
theRealRizeo: Added escapes to the urls

includes/controllers/pages/class-manage-listings.php
theRealRizeo: Updated the comment and removed the `since` tag as there is no reference on the changes or when the function was added.

stephywells (approved):
Looks great! Thank you!!

PR #69, opened 2021-11-04
Hide payment form when not recurring and when price is 0

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4993

- If the plan is not recurring and an discount code of 100% is applied, the payment options are hidden
- If the plan is recurring but there is a discount code of 100%, the payment options will be shown
- If it is full price, the payment options will be shown

### DIFF
--- a/includes/controllers/pages/class-checkout.php
+++ b/includes/controllers/pages/class-checkout.php
@@ -178,7 +178,7 @@ class WPBDP__Views__Checkout extends WPBDP__View {
         $pay_now = ( $this->payment->has_item_type( 'recurring_plan' ) || $this->payment->amount > 0 );
         $checkout_form .= sprintf(
             '<div class="wpbdp-checkout-submit"><input type="submit" value="%s" /></div>',
-            $pay_now ? _x( 'Pay Now', 'checkout', 'business-directory-plugin' ) : _x( 'Complete', 'checkout', 'business-directory-plugin' )
+            $this->payment->show_payment_options() ? _x( 'Pay Now', 'checkout', 'business-directory-plugin' ) : _x( 'Complete', 'checkout', 'business-directory-plugin' )
         );
 
         return $checkout_form;
--- a/includes/controllers/pages/class-checkout.php
+++ b/includes/controllers/pages/class-checkout.php
@@ -177,7 +177,7 @@ class WPBDP__Views__Checkout extends WPBDP__View {
         // $checkout_form .= wpbdp_capture_action( 'wpbdp_checkout_form_bottom', $this->payment );
         $checkout_form .= sprintf(
             '<div class="wpbdp-checkout-submit"><input type="submit" value="%s" /></div>',
-            $this->payment->show_payment_options() ? esc_html__( 'Pay Now', 'business-directory-plugin' ) : esc_html__( 'Complete', 'business-directory-plugin' )
+            $this->payment->show_payment_options() ? esc_attr__( 'Pay Now', 'business-directory-plugin' ) : esc_attr__( 'Complete', 'business-directory-plugin' )
         );
 
         return $checkout_form;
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
Checkouts work as described when a discount is applied.

Non-recurring:
Screenshot 2021-11-04 at 14-54-21 BD QA – Just another WordPress site

Recurring:
Screenshot 2021-11-04 at 15-00-03 BD QA – Just another WordPress site

includes/controllers/pages/class-checkout.php
```diff
@@ -175,7 +175,11 @@ private function checkout_form() {
         // $checkout_form .= wpbdp_capture_action( 'wpbdp_checkout_form_top', $this->payment );
         $checkout_form .= $this->gateway->render_form( $this->payment, $this->errors );
         // $checkout_form .= wpbdp_capture_action( 'wpbdp_checkout_form_bottom', $this->payment );
-        $checkout_form .= '<div class="wpbdp-checkout-submit"><input type="submit" value="' . _x( 'Pay Now', 'checkout', 'business-directory-plugin' ) . '" /></div>';
+        $pay_now = ( $this->payment->has_item_type( 'recurring_plan' ) || $this->payment->amount > 0 );
```
stephywells, line 178: Looks like this line isn't used

includes/controllers/pages/class-checkout.php
```diff
@@ -175,7 +175,11 @@ private function checkout_form() {
...
+        $checkout_form .= sprintf(
+            '<div class="wpbdp-checkout-submit"><input type="submit" value="%s" /></div>',
+            $this->payment->show_payment_options() ? _x( 'Pay Now', 'checkout', 'business-directory-plugin' ) : _x( 'Complete', 'checkout', 'business-directory-plugin' )
```
stephywells, line 181: This is missing sanitizing

stephywells (changes requested):
Nice progress on this!

includes/controllers/pages/class-checkout.php
theRealRizeo: Not sure about what should be sanitized here, but text has been changed to use `esc_html__`

stephywells: Yep, exactly. The translations are output that should always be sanitized too.

includes/controllers/pages/class-checkout.php
```diff
@@ -175,7 +175,10 @@ private function checkout_form() {
         // $checkout_form .= wpbdp_capture_action( 'wpbdp_checkout_form_top', $this->payment );
         $checkout_form .= $this->gateway->render_form( $this->payment, $this->errors );
         // $checkout_form .= wpbdp_capture_action( 'wpbdp_checkout_form_bottom', $this->payment );
-        $checkout_form .= '<div class="wpbdp-checkout-submit"><input type="submit" value="' . _x( 'Pay Now', 'checkout', 'business-directory-plugin' ) . '" /></div>';
+        $checkout_form .= sprintf(
+            '<div class="wpbdp-checkout-submit"><input type="submit" value="%s" /></div>',
+            $this->payment->show_payment_options() ? esc_html__( 'Pay Now', 'business-directory-plugin' ) : esc_html__( 'Complete', 'business-directory-plugin' )
```
stephywells, line 180: Actually... These need to use esc_attr instead of esc_html since it's inside of an HTML attribute.

`$this->payment->show_payment_options() ? esc_attr__( 'Pay Now', 'business-directory-plugin' ) : esc_attr__( 'Complete', 'business-directory-plugin' )`

theRealRizeo: Thanks, this has been updated :)

includes/models/class-payment.php
```diff
@@ -287,7 +287,7 @@ public function get_checkout_url( $force_http = false ) {
         $url = wpbdp_url( 'checkout', $this->payment_key );
 
         if ( ! $force_http && ! is_ssl() ) {
-            $url = set_url_scheme( $url, 'https' );
+            //$url = set_url_scheme( $url, 'https' );
```
stephywells, line 290: Why is this commented out?

stephywells (changes requested):
This is close!

includes/models/class-payment.php
theRealRizeo: Forgot to comment it back, its for my local tests

stephywells (approved):
Thanks Paul! Looks good now!

PR #70, opened 2021-11-09
Remove external FontAwesome style asset loading

theRealRizeo:
Related : https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4994

This notifies the user to install the official [Font Awesome](https://wordpress.org/plugins/font-awesome) plugin.
Please advise if we should make this process easier for the user


stephywells (changes requested):
@theRealRizeo can you do this in a way that won't break sites that are using the existing setting? Changing the name of the setting will add a lot of extra complication.

Also, since the setting isn't being used, we don't need it at all.

So, here's what I'm thinking... We can switch the existing setting to 'type' => 'hidden' and the default to empty. This way, we know which sites had it turned on.

Then, we should probably add an admin notice with a link to install the Font Awesome plugin if the setting was turned on. This message can be dismissed too. If dismissed, it would change the setting.

theRealRizeo:
Admin notice added and original setting set to hidden

includes/admin/class-admin.php
```diff
@@ -1162,6 +1173,33 @@ private function check_deprecation_warnings() {
...
 
+		/**
+		 * Check old inline fontawesome setting.
+		 * If it was enabled by default, we show an admin notice.
+		 * On dismiss we keep the setting.
+		 *
+		 * @since x.x
+		 */
+		private function check_font_awesome_setting() {
+			if ( ! current_user_can( 'administrator' ) ) {
```
stephywells, line 1184: Let's use install_plugins here.

We could also include `! WPBDP_App_Helper::is_bd_page()` in the same line too.

includes/admin/class-admin.php
```diff
@@ -1162,6 +1173,33 @@ private function check_deprecation_warnings() {
...
+			if ( ! WPBDP_App_Helper::is_bd_page() ) {
+				return;
+			}
+
+			if ( ! wpbdp_get_option( 'enqueue-fontawesome-styles', false ) ) {
+				return;
+			}
+
+			$this->messages[] = array(
+				str_replace( array( '[', ']' ), array( '<a href="https://wordpress.org/plugins/font-awesome" target="_blank" rel="noopener nofollow">', '</a>' ), _x( 'We changed how FontAwesome is integrated in Business Directory. Go [here] to install the officle "FontAwesome styles" plugin.', 'admin', 'business-directory-plugin' ) ),
```
stephywells, line 1197: I know str_replace is used throughout this plugin for translations like this, but it's not the right way. It should use sprintf

includes/admin/class-admin.php
```diff
@@ -1162,6 +1173,33 @@ private function check_deprecation_warnings() {
...
+				'error dismissible',
+				array( 'dismissible-id' => 'fontawesome_dismissed' ),
```
stephywells, line 1199: This name is making the function name super long. Can we condense this down to just 'fa' or something?

includes/admin/class-admin.php
```diff
@@ -1133,6 +1135,15 @@ public function ajax_dismiss_notification_server_requirements() {
             set_transient( 'wpbdp_server_requirements_warning_dismissed', true, WEEK_IN_SECONDS );
         }
 
+		/**
+		 * Set fontawesome notice as dismissed.
+		 *
+		 * @since x.x
+		 */
+		public function ajax_dismiss_notification_fontawesome_dismissed() {
+			wpbdp_set_option( 'enqueue-fontawesome-styles', false );
```
stephywells, line 1144: We still need nonce and permissions checks here since it's an ajax call.

stephywells (changes requested):

includes/admin/class-admin.php
garretlaxton: There's a minor typo here, the word `official` is accidentally spelled `officle`.

garretlaxton (commented):
The notice banner doesn't disappear once the FA plugin is installed. Other than that, this works fine.
Screenshot 2021-11-09 at 22-27-49 Settings ‹ BD QA — WordPress

includes/admin/class-admin.php
theRealRizeo: Nonce are handled in the function `ajax_dismiss_notification` that is used on the plugins dismiss messages. I'll add permission check on it

includes/admin/class-admin.php
theRealRizeo: 'FA' might not be a name common to most users

theRealRizeo:
@garretlaxton that should not show now, thanks for pointing that out

includes/admin/class-admin.php
theRealRizeo: I've changed it to `sprintf` and made the line shorter

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -350,13 +350,42 @@ private static function settings_general() {
...
+		if ( ! is_plugin_active( 'font-awesome/index.php' ) || ( ! class_exists( '\FortAwesome\FontAwesome_Loader' ) ) ) {
+			wpbdp_register_setting(
+				array(
+					'id'    => 'fontawesome-enabled',
+					'name'  => _x( 'FontAwesome styles?', 'settings', 'business-directory-plugin' ),
+					'desc'  => wp_kses_post(
+						sprintf(
+							/* translators: %s: addon name */
+							'<span>' . __( 'Install %1$s for FontAwesome styles?', 'business-directory-plugin' ) . '</span>',
+							'<a href="https://wordpress.org/plugins/font-awesome" target="_blank" rel="noopener nofollow">Font Awesome</a> ',
```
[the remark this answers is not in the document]
theRealRizeo: For the field icons it might be better to leave it in the settings

includes/admin/class-admin.php
stephywells: I'm specifically referring to "fontawesome_dismissed" since it's creating a hook like "wpbdp_admin_ajax_dismiss_notification_fontawesome_dismissed". Include "dismissed" is overkill and making for really long function names.

includes/admin/settings/class-settings-bootstrap.php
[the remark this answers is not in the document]
stephywells: Can you tell me why you would like it in the settings?

includes/admin/class-admin.php
```diff
@@ -1162,6 +1175,39 @@ private function check_deprecation_warnings() {
...
+            if ( is_plugin_active( 'font-awesome/index.php' ) && ( class_exists( '\FortAwesome\FontAwesome_Loader' ) ) ) {
+                wpbdp_set_option( 'enqueue-fontawesome-styles', false );
+                return;
+            }
+
+			$this->messages[] = array(
+                sprintf( 
+                    _x( 'We changed how %1$s is integrated into Business Directory. Go %2$shere%3$s to install the official "%1$s styles" plugin.', 'admin', 'business-directory-plugin' ),
+                    'FontAwesome',
+                    '<a href="https://wordpress.org/plugins/font-awesome" target="_blank" rel="noopener nofollow">',
```
stephywells, line 1203: This url should be updated too.

stephywells (changes requested):
Can you please review the PHPCS and syntax checks on this PR?

Also, please check the tabs vs spaces on every line that has been changed. Please don't apply it to the entire files, but only the lines that have changed to keep the diff low.

includes/admin/settings/class-settings-bootstrap.php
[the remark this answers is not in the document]
theRealRizeo: The fonts will not only be used in the social field type when the field icons are implemented. This will be a common setting to be used in the field icons display and would work better as a general setting

includes/admin/class-admin.php
theRealRizeo: Changes done :)

includes/admin/settings/class-settings-bootstrap.php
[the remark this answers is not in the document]
stephywells: The problem is, the spot it's in right now just doesn't make sense. If we include it in the settings, it should be at the bottom of the appearance tab.

includes/admin/class-admin.php
stephywells: Thanks! Looks like the tabs vs spaces are still messed up throughout all these changes.

includes/admin/settings/class-settings-bootstrap.php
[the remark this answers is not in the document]
theRealRizeo: All changed to tabs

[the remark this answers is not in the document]
theRealRizeo: Also settings moved

garretlaxton:
The notification doesn't appear when the plugin isn't installed. 
Screenshot 2021-11-11 at 13-22-40 Settings ‹ BD QA — WordPress

When the plugin is installed, the setting disappears from the appearance tab. I'm not sure if that's expected, but it may confuse some users.
Screenshot 2021-11-11 at 13-28-08 Settings ‹ BD QA — WordPress

theRealRizeo:
Yes, once active the notice will not show. It would be a good idea to show something to the user as they might be confused and bring up more questions

garretlaxton:
Maybe changing the notice text to something like `Installed` would be enough.

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -350,16 +350,18 @@ private static function settings_general() {
...
-            )
-        );
+
+		wpbdp_register_setting(
+			array(
+				'id'      => 'enqueue-fontawesome-styles',
+				'type'    => 'hidden',
+				'class'   => 'hidden',
+				'name'    => _x( 'Enqueue Business Directory\'s FontAwesome styles?', 'settings', 'business-directory-plugin' ),
+				'desc'    => _x( 'This helps to prevent conflicts with other plugins that already do this. Disable this only if you\'re having an issue with FontAwesome icons and have performed a conflict test to validate this is a multiple styles enqueueing issue.', 'settings', 'business-directory-plugin' ),
```
stephywells, line 360: Are the name and description on the hidden field used anywhere now? Or can they be removed?

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -915,6 +917,23 @@ private static function settings_appearance() {
 
 		WPBDP_Admin_Education::add_tip_in_settings( 'table', 'themes' );
 
+        if ( ! is_plugin_active( 'font-awesome/index.php' ) || ( ! class_exists( '\FortAwesome\FontAwesome_Loader' ) ) ) {
+			wpbdp_register_setting(
+				array(
+					'id'    => 'fontawesome-enabled',
+					'name'  => _x( 'FontAwesome styles?', 'settings', 'business-directory-plugin' ),
```
stephywells, line 924: Let's change this to `__( 'Font Awesome', 'business-directory-plugin' ),` (It's two words)

includes/admin/class-admin.php
```diff
@@ -1162,6 +1175,39 @@ private function check_deprecation_warnings() {
...
+
+			if ( is_plugin_active( 'font-awesome/index.php' ) && ( class_exists( '\FortAwesome\FontAwesome_Loader' ) ) ) {
+				wpbdp_set_option( 'enqueue-fontawesome-styles', false );
+				return;
+			}
+
+			$this->messages[] = array(
+				sprintf(
+					_x( 'We changed how %1$s is integrated into Business Directory. Go %2$shere%3$s to install the official "%1$s styles" plugin.', 'admin', 'business-directory-plugin' ),
+					'FontAwesome',
```
stephywells, line 1202: I don't see a reason to keep this out of the translatable string since we are translating it in other spots. Also, please make sure we are using "Font Awesome" consistently with a space.

includes/admin/class-admin.php
```diff
@@ -1162,6 +1175,39 @@ private function check_deprecation_warnings() {
...
+		private function check_font_awesome_setting() {
+			if ( ! current_user_can( 'install_plugins' ) || ! WPBDP_App_Helper::is_bd_page() ) {
+				return;
+			}
+
+			if ( ! wpbdp_get_option( 'enqueue-fontawesome-styles', false ) ) {
+				return;
+			}
+
+			if ( is_plugin_active( 'font-awesome/index.php' ) && ( class_exists( '\FortAwesome\FontAwesome_Loader' ) ) ) {
```
stephywells, line 1194: Is there a reason for both of these checks? Or is one enough?

includes/admin/class-admin.php
```diff
@@ -1162,6 +1175,39 @@ private function check_deprecation_warnings() {
...
+				wpbdp_set_option( 'enqueue-fontawesome-styles', false );
+				return;
+			}
+
+			$this->messages[] = array(
+				sprintf(
+					_x( 'We changed how %1$s is integrated into Business Directory. Go %2$shere%3$s to install the official "%1$s styles" plugin.', 'admin', 'business-directory-plugin' ),
```
stephywells, line 1201: How about this?
`__( 'Good news! Business Directory Plugin now integrates with the official Font Awesome plugin. %1$sInstall Font Awesome now%2$s.', 'business-directory-plugin' ),`

includes/admin/class-admin.php
```diff
@@ -1144,11 +1157,11 @@ public function check_setup() {
             if ( wpbdp_get_option( 'require-login' )
                 && ! get_option( 'users_can_register' )
                 && ! get_user_meta( get_current_user_id(), 'wpbdp_notice_dismissed[registration_disabled]', true ) ) {
-                    $this->messages[] = array(
-                        str_replace( array( '[', ']' ), array( '<a href="' . esc_url( admin_url( 'options-general.php' ) ) . '">', '</a>' ), _x( 'We noticed you want your Business Directory users to register before posting listings, but Registration for your site is currently disabled. Go [here] and check "Anyone can register".', 'admin', 'business-directory-plugin' ) ),
-                        'error dismissible',
-                        array( 'dismissible-id' => 'registration_disabled' ),
-                    );
+					$this->messages[] = array(
+						sprintf( _x( 'We noticed you want your Business Directory users to register before posting listings, but Registration for your site is currently disabled. Go %1$shere%2$s and check "Anyone can register".', 'admin', 'business-directory-plugin' ), '<a href="' . esc_url( admin_url( 'options-general.php' ) ) . '">', '</a>' ),
```
stephywells, line 1161: For PHPCS, sprintf( should be on its own line for better readability. This is really long to keep all on one line.

stephywells (changes requested):
This is getting close!

stephywells:
> Maybe changing the notice text to something like `Installed` would be enough.

I agree that it might be confusing if it looks like there's a setting, the FA plugin is installed, and then the "setting" disappears. The problem is that there aren't any settings, and there are a bunch of different plugins and themes that load Font Awesome. So if they don't need the plugin we're recommending, this just sits and takes up a bunch of space in the settings.

There's got to be a better way of doing this. Any suggestions about where to let people know that Business Directory uses Font Awesome icons?

garretlaxton:
> There's got to be a better way of doing this. Any suggestions about where to let people know that Business Directory uses Font Awesome icons?

Maybe a one-time notification when the user visits the appearance tab, they can see and close it?

theRealRizeo:
Maybe some small text with a link to a blog post on why it was changed?

theRealRizeo:
Pushed an update:

- If plugin is not installed, it will show the message to alert the user to install it
- If plugin is installed, we show the same message without the notification to install the plugin.
- If the message is dismissed, we assume the user know the changes to the plugin and has accepted this will not be shown

includes/admin/class-admin.php
theRealRizeo: Second check has been changed to notify the user who has the plugin active, and did not dismiss the notice before, of the changes

garretlaxton (approved):
I think this is good enough.

includes/admin/class-admin.php
```diff
@@ -1162,6 +1179,42 @@ private function check_deprecation_warnings() {
...
+					'notice dismissible',
+					array( 'dismissible-id' => 'fa_dismiss' ),
+				);
+				return;
+			}
+
+			$this->messages[] = array(
+				sprintf(
+					__( 'Good news! Business Directory Plugin now integrates with the official Font Awesome plugin. %1$sInstall Font Awesome now%2$s.', 'business-directory-plugin' ),
+					'<a href=" ' . esc_url( admin_url( 'plugin-install.php?s=fontawesome&tab=search&type=author' ) ) . ' " target="_blank" rel="noopener nofollow">',
```
stephywells, line 1210: I think we either need to not open this in a new tab, or make the link trigger the dismissal. Right now, when you click it, go install, and come back, the message is still there.

includes/admin/class-admin.php
```diff
@@ -1162,6 +1179,42 @@ private function check_deprecation_warnings() {
...
+				return;
+			}
+
+			if ( ! wpbdp_get_option( 'enqueue-fontawesome-styles', false ) ) {
+				return;
+			}
+
+			if ( is_plugin_active( 'font-awesome/index.php' ) ) {
+				$this->messages[] = array(
+					__( 'Good news! Business Directory Plugin now integrates with the official Font Awesome plugin.', 'business-directory-plugin' ),
```
stephywells, line 1200: This message doesn't look very helpful to me. Let's remove it.

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -915,6 +915,23 @@ private static function settings_appearance() {
 
 		WPBDP_Admin_Education::add_tip_in_settings( 'table', 'themes' );
 
+        if ( ! is_plugin_active( 'font-awesome/index.php' ) || ( ! class_exists( '\FortAwesome\FontAwesome_Loader' ) ) ) {
+			wpbdp_register_setting(
+				array(
+					'id'    => 'fontawesome-enabled',
+					'name'  => __( 'Font Awesome', 'business-directory-plugin' ),
```
stephywells, line 922: Let's get rid of this name line for education. It makes it looks like there will be a setting when there isn't.

includes/admin/class-admin.php
```diff
@@ -1144,11 +1157,15 @@ public function check_setup() {
...
                 && ! get_option( 'users_can_register' )
                 && ! get_user_meta( get_current_user_id(), 'wpbdp_notice_dismissed[registration_disabled]', true ) ) {
-                    $this->messages[] = array(
-                        str_replace( array( '[', ']' ), array( '<a href="' . esc_url( admin_url( 'options-general.php' ) ) . '">', '</a>' ), _x( 'We noticed you want your Business Directory users to register before posting listings, but Registration for your site is currently disabled. Go [here] and check "Anyone can register".', 'admin', 'business-directory-plugin' ) ),
-                        'error dismissible',
-                        array( 'dismissible-id' => 'registration_disabled' ),
-                    );
+					$this->messages[] = array(
+						sprintf(
+                            __( 'We noticed you want your Business Directory users to register before posting listings, but Registration for your site is currently disabled. Go %1$shere%2$s and check "Anyone can register".', 'business-directory-plugin' ),
```
stephywells, line 1162: Please check tabs on these lines

includes/admin/class-admin.php
```diff
@@ -1162,6 +1179,42 @@ private function check_deprecation_warnings() {
...
+		private function check_font_awesome_setting() {
+			if ( ! current_user_can( 'install_plugins' ) || ! WPBDP_App_Helper::is_bd_page() ) {
+				return;
+			}
+
+			if ( ! wpbdp_get_option( 'enqueue-fontawesome-styles', false ) ) {
+				return;
+			}
+
+			if ( is_plugin_active( 'font-awesome/index.php' ) ) {
```
stephywells, line 1198: Let's switch all these checks to be more universal like we did in the other PR. We should check the styles instead of a single plugin.

stephywells (changes requested):
This is getting close!

includes/admin/class-admin.php
theRealRizeo: I've made a change where on click, the notice will be dismissed, it wont open in a new tab. The notice will not show again other than in the Appearance tab in settings

includes/admin/class-admin.php
theRealRizeo: Moved the checks to the helper class

includes/admin/class-admin.php
theRealRizeo: Message removed

assets/js/admin.js
```diff
@@ -557,6 +557,14 @@ jQuery(function($) {
 
                 dismissNotice( $notice, dismissible_id, nonce );
             } );
+
+        $( document ).on( 'click', '.wpbdp-admin-silent-dismiss', function( e ) {
```
stephywells, line 561: Would it work to add the "notice-dismiss" class on the link in the message instead of adding extra js?

includes/helpers/class-app.php
```diff
@@ -345,6 +345,21 @@ public static function kses( $value, $allowed = array() ) {
...
+	/**
+	 * Check if font awesome is enabled.
+	 * This checks if the plugin is active or the styles are loaded.
+	 *
+	 * @since x.x
+	 *
+	 * @return bool
+	 */
+	public static function is_font_awesome_enabled() {
+		if ( is_plugin_active( 'font-awesome/index.php' ) || wp_style_is( 'font-awesome', 'enqueued' ) ) {
```
stephywells, line 357: Can we slim this down to:
`return is_plugin_active( 'font-awesome/index.php' ) || wp_style_is( 'font-awesome', 'enqueued' );`

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -915,6 +915,22 @@ private static function settings_appearance() {
 
 		WPBDP_Admin_Education::add_tip_in_settings( 'table', 'themes' );
 
+        if ( ! WPBDP_App_Helper::is_font_awesome_enabled() ) {
```
stephywells, line 918: Please check tabs

includes/admin/settings/class-settings-bootstrap.php
```diff
@@ -915,6 +915,22 @@ private static function settings_appearance() {
...
+			wpbdp_register_setting(
+				array(
+					'id'    => 'fontawesome-enabled',
+					'desc'  => wp_kses_post(
+						sprintf(
+							'<span>' . __( 'Install %s for Font Awesome styles?', 'business-directory-plugin' ) . '</span>',
```
stephywells, line 924: Let's use a message that's a bit more descriptive like this:
Use icons in your directory listings and custom fields: Install %s

stephywells (commented):
Mostly just small issues now!

assets/js/admin.js
theRealRizeo: Hey Steph, it wouldn't work mainly because the `notice-dismiss` checks for `parent` which is the direct parent of the button. The link sits within  a parent element `p` hence the new js function uses `closest` https://stackoverflow.com/questions/10499435/parent-vs-closest

Another issue is the class `notice-dismiss` is styled to work as a button. Assigning it to the link has the following styling effect
Screenshot 2021-11-20 092112

includes/admin/settings/class-settings-bootstrap.php
theRealRizeo: This is much better, thanks :)

assets/js/admin.js
stephywells: Gotcha. Thanks for checking. 

Maybe we could reduce the duplication here a bit although it already existed? Maybe you could look into it to see if these are both being used? It's possible one is for a version of WordPress we no longer support.

That would probably help to move some of the duplication into the dismissNotice function.

theRealRizeo: Yes, agreed :) Will check the code for this

theRealRizeo: Done some changes to combine the js into one. Added a callback parameter that is used to send to the next page after the background dismiss action is passed. This is mainly because the ajax action is not completed before the link opens the new page. It ensure the ajax action is completed before sending to the new page

stephywells:
Hey @garretlaxton would you mind checking this once more? I made some pretty major changes with moving and cleaning up this code. I also updated it to check for Better Font Awesome plugin as well.

stephywells (approved):

PR #74, opened 2021-11-13
Fix `Never` expire status from admin

theRealRizeo:
When `never` is selected from the date picker in the admin side of the listing (backend admin), the listing should set to never. Right now it shows as expired which is not the expected outcome

### DIFF
--- a/includes/admin/controllers/class-admin-listings.php
+++ b/includes/admin/controllers/class-admin-listings.php
@@ -600,7 +600,7 @@ class WPBDP_Admin_Listings {
 
         // Update plan attributes.
         $row                    = array();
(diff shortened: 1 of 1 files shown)


stephywells (commented):
@theRealRizeo did you test this PR? This doesn't look like it would solve the problem.

theRealRizeo:
@stephywells I tested it multiple times over the weekend. When the time is null, it sets to never expire

garretlaxton:
This fixed it for me. Also made sure you can still set listings as expired.
Screenshot 2021-11-15 at 14-47-35 Edit Listing ‹ BD QA — WordPress

Can we please prioritize this release as I've gotten quite a few tickets about it today.

garretlaxton (approved):

stephywells:
Sure! We can plan on a release tomorrow. Thank you both!

PR #77, opened 2021-11-17
Include minification or premium file js

theRealRizeo:
Small update to include the `js` in the `business-directory-premium` addon` in the grunt task


Gruntfile.js
```diff
@@ -305,6 +305,7 @@ module.exports = function( grunt ) {
   grunt.wpbdp.registerModule({path: '../business-directory-stripe', js: ['resources/*.js'], i18n: true});
   grunt.wpbdp.registerModule({path: '../business-directory-zipcodesearch', less: ['resources/*.css'], js: ['resources/*.js'], i18n: true});
   grunt.wpbdp.registerModule({path: '../business-directory-premium', less: ['resources/*.css'], js: [], i18n: {'textDomain':'wpbdp-pro'}});
+  grunt.wpbdp.registerModule({path: '../business-directory-premium', js: ['resources/*.js'], i18n: true});
```
stephywells, line 308: Since this module is already included in line 307, we should update that line instead of adding another..

stephywells (changes requested):

Gruntfile.js
theRealRizeo: Oh, yes :) Agreed

stephywells (approved):
Perfect! Thanks!

PR #78, opened 2021-11-18
Complete never expiring listings set to publish

theRealRizeo:
This is in reference to https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4998

When a listing that is set to never expire is approved by the admin, by clicking the publish button, the listing status should go direct to publish. At the moment it requires 2 clicks : first status is updated from "Pending Review" to "Draft". Then from "Draft" to "Publish". This skips the two steps

### DIFF
--- a/includes/admin/controllers/class-admin-listings.php
+++ b/includes/admin/controllers/class-admin-listings.php
@@ -611,10 +611,10 @@ class WPBDP_Admin_Listings {
 		}
 
 		// Check if the status needs to be changed.
-		if ( 'expired' == $listing->get_status() ) {
-			if ( ! $row['expiration_date'] || $not_expired ) {
-				$listing->get_status( true, true );
-			}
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
Works great, thanks Paul! Not sure how this one slipped past.

We should prioritize this as well since we are getting tickets for this.

stephywells (commented):
This looks like a lot of workaround and extra code to fix the issue here. In doing so without any unit tests, this is bound to break something else. I've got another PR incoming with much fewer changes to fix the new issue. I don't feel comfortable with a change this big that should be a one-liner right before Black Friday. We can't introduce new bugs right now.

theRealRizeo:
Some code changes is mainly to simplify existing code. There is only 1 new function and a check to update the listing post status

stephywells:
New PR here:
https://github.com/Strategy11/business-directory-plugin/pull/79

theRealRizeo:
Works great :) Closing this

PR #80, opened 2021-11-18
Fee plan update action trigger image limit notice

theRealRizeo:
This is related to https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4981

- If a fee plan is updated, there will be an additional link in the admin notice to update the listing that share the same fee plan
- This link will only show if the fee plan has listings associated with it
- Adds a new cache db call function in the class `WPBDP__Fee_Plan` called `count_listings()` to check the number of listings associated with the fee plan
- This also introduces the possibility of adding a column in the "Fee Plans" view for number of listings the current plan is associated with
- There is also a new JS function that uses the already existing admin ajax calls


garretlaxton (commented):
This works good, the only thing is if you don't click to update the image limits, then you don't have a choice to do so later. Not sure if there's anything we can do about that.

theRealRizeo:
Since its a one-time action, we want to only show it once on update. But I guess having a button within the edit screen of the plan, with a notice of the repercussions of clicking the said button, to update listing images can be a good idea

stephywells (approved):
Nice job on this @theRealRizeo ! Thanks for testing it so well!

PR #83, opened 2021-11-22
Excerpt view default theme

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4986

This fixes the excerpt view for the default theme to match the field names when updated.
The fix makes sure the default theme excerpt view layout matches that of the template file


stephywells (changes requested):
Thanks @theRealRizeo ! Instead of adding duplicate code, can we use an include here instead?

theRealRizeo:
@stephywells agreed :) I've moved the contents of the excerpt content to the directory `templates\parts\excerpt-details.tpl.php` . This allows including the same repetitive content in the default plugin template and the theme template

stephywells (changes requested):
Looks good! Let's go ahead and change the reference to wpbdp_x_part( 'excerpt_content' ) also. Maybe we should add a note that the excerpt_content file is deprecated, but not add a notice yet?

theRealRizeo:
@stephywells I have checked other BD themes and that may change the default `themes/default/templates/excerpt_content.tpl.php` template file loading for the default theme and other themes such as `modern-business` as each have their own custom `excerpt_content.tpl.php`  and `excerpt.tpl.php` (in regards to modern business theme). This might bring about un-wanted changes in appearance

garretlaxton:
Changes come through great. Nice job!

The only thing I'm seeing is the title shown twice in the excerpt view. You can see this right below the ratings.
Screenshot 2021-11-22 at 22-06-54 BD QA – Just another WordPress site

theRealRizeo:
@garretlaxton thanks for catching that :) Pushed a fix for it

garretlaxton (approved):
Looks good!
Screenshot 2021-11-23 at 13-59-48 BD QA – Just another WordPress site

stephywells (changes requested):
After looking at this a bit more, I think it changes it a lot more than it should. It looks like we should back out all these changes and just fix the 'Address' line.

Or, we could add the formatted address to the excerpt template file, since it would still be nice to get rid of extra template files in themes.

theRealRizeo:
I think for the `default` theme the goal was to set the `Address` as a combination of all the address fields giving a more compact look for the theme as opposed to other themes. `Modern Business Theme` also does not show the `Address` layout or other field labels. Perhaps we just need better documentation of what each theme does?

theRealRizeo:
Looked at the code again and made some changes to get the `Address` label from the corresponding field. Added a helper function to also reflect this.  @garretlaxton please check, this reverts previous changes and only modifies the `Address` label to that of the field

garretlaxton:
The field label changes to the correct value, however, the label doesn't have any styling. The text is black and the field icon is missing.
Screenshot 2021-11-29 at 19-39-02 BD QA – Just another WordPress site

Here is what it should look like:
Screenshot 2021-11-29 at 19-39-21 Andy Agent – BD QA

theRealRizeo:
@garretlaxton thanks :) Fixed. Please check again

garretlaxton:
There is an extra colon after the label, probably due to having a field icon and label displaying. 
Screenshot 2021-11-30 at 14-54-58 BD QA – Just another WordPress site

Also this field doesn't follow it's correct order always displaying first. My ratings and brief description fields should be above the address as seen here.
Screenshot 2021-11-23 at 13-59-48 BD QA – Just another WordPress site

theRealRizeo:
> There is an extra colon after the label, probably due to having a field icon and label displaying. Screenshot 2021-11-30 at 14-54-58 BD QA – Just another WordPress site
> 
> Also this field doesn't follow it's correct order always displaying first. My ratings and brief description fields should be above the address as seen here. Screenshot 2021-11-23 at 13-59-48 BD QA – Just another WordPress site

@garretlaxton thanks, this has been fixed :)

garretlaxton:
The sort order still isn't correct. My address field should display at the bottom according to my order, but is on top.
Screenshot 2021-12-01 at 14-30-22 Form Fields ‹ BD QA — WordPress
Screenshot 2021-12-01 at 14-33-00 BD QA – Just another WordPress site

And one last thing, when you select `Hide Label` the label still shows.
Screenshot 2021-12-01 at 14-29-36 Form Fields ‹ BD QA — WordPress

theRealRizeo:
@garretlaxton please check and confirm if other themes follow these rules : Hide label and Sort order.
Default Theme has a specific order for the fields, the first is `Address`
Modern business for example, does not show the label in most views. I think the extra changes you have pointed out should be moved to a different issue as we will need to define what the various themes show and how they are to interact with the different field settings.

garretlaxton:
Switching from the `default theme` to `no theme` does properly show the correct order. I wasn't aware the default theme had a separate order. I personally think that should be changed to allow a custom order, but that's a separate issue as mentioned.

garretlaxton (approved):

stephywells (changes requested):
Oops! I just realized I left comments pending on this a while back.

stephywells (approved):
I made a few updates since Paul is out for the day.

This has been updated to include the address in the right order. Thanks getting this most of the way Paul, and thanks for the testing @garretlaxton !

stephywells:
This has been update to include the address in the right order. Thanks getting this most of the way Paul, and thanks for the testing @garretlaxton !

theRealRizeo:
Thanks @stephywells :) This came in at around 1:30 am my local time

PR #84, opened 2021-11-24
Test cases

theRealRizeo:
More test case updates. This moves files from the test updates in https://github.com/Strategy11/business-directory-plugin/pull/65 to this branch and adds other test cases


stephywells:
Thanks for working on this @theRealRizeo !

I'm seeing there are tests an no assertions.
https://github.com/Strategy11/business-directory-plugin/runs/4434758974?check_suite_focus=true#step:9:39

Are you seeing this in your local testing too? Without any assertions, nothing is actually getting tested.

theRealRizeo:
Hey @stephywells I will check those tests

theRealRizeo:
Fixing this then will re-open

theRealRizeo:
Hey @stephywells did some deeper digging on Acceptance tests. For the tests to handle the "click", "inputs", etc it requires a sandbox site to perform these actions. In this case, preferably one where a deployment of the current changes to be tested are done. Do we have a resource where this can be setup?

tests/wpunit/Notification/AbandonedPaymentTest.php
```diff
@@ -0,0 +1,58 @@
...
+namespace Notification;
+
+use WPBDP\Tests\WPUnitTestCase;
+use WPBDP__Abandoned_Payment_Notification;
+use WPBDP__Settings;
+
+/**
+ * Tests for the Notifications.
+ */
+class AbandonedPaymentTest extends WPUnitTestCase {
```
stephywells, line 14: We should actually move this set of tests into the BD Premium plugin

stephywells (commented):
Is this ready? I'm not seeing the tests running here in Github...

theRealRizeo:
Almost ready. Will remove acceptance tests and move them all to wpunit

.github/workflows/codeception.yml
```diff
@@ -51,7 +51,13 @@ jobs:
         run: |
           cd "/tmp/wordpress/src/wp-content/plugins/${GITHUB_REPOSITORY#*/}"
           pwd
-          php vendor/bin/codecept run
+          php vendor/bin/codecept run wpunit Data/ExporterTest.php
```
stephywells, line 54: Looks like this got a lot more complicated.... Can we do this in one line instead of one per file?

theRealRizeo: Hey @stephywells before I set it to run all, but it would always return an error `Warning: mysqli_query(): Couldn't fetch mysqli` . I looked around and found this https://githubmemory.com/repo/lucatume/wp-browser/issues/497 . They will work well when singular and not in parallel

theRealRizeo: Fixed it, works now as a single suite

theRealRizeo: Error for MySQL was when trying to run all different tests at once (acceptance and wpunit)

stephywells: This is looking much better now. As long as you are positive a failed test will fail here in Github, we can go ahead and merge this.

theRealRizeo: Yes, I'm positive. I have some failed tests locally for testing purposes and the results fail like if its phpunit

PR #85, opened 2021-12-01
Hide scroll options for categories

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5001

- This hides the `Number of options visible without scrolling` option when `MultiSelect` Field association and `Category` field type are selected.
- Category shows options with `select2` and does not have this option
- Other `MultiSelect` options that are not categories will have the option
- Change uses Javascript to show and hide the field container with this setting
- There is an additional class added to the section wrapper `wpbdp-fieldsettings-{fieldtype}` which in this case is `wpbdp-fieldsettings-multiselect` that is used to target and show/hide


garretlaxton (approved):
Looks good! The field-specific settings are now hidden when multi-select and category are both selected.
Screenshot 2021-12-01 at 17-27-56 Form Fields ‹ BD QA — WordPress

stephywells:
@theRealRizeo I went ahead and redid this since you're out sick. It turns out, there's a much simpler way of doing this.

theRealRizeo:
Thanks @stephywells :)

PR #88, opened 2021-12-08
Remove automattic image resizing on every view.

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5002

- Removes the automatic check and resize of listing images for every view. This only works for new image uploads
- Thumbnail creation is only done after the image is uploaded.
- Introduces CSS override to set the max and min width and height of the image in the frontend


garretlaxton (approved):
Looks good!

stephywells:
@garretlaxton @theRealRizeo I just made some pretty big changes with the way the file uploads work, so WP handles all the image sizes. But the problem is, I don't have GD library set up and don't have the time to reinstall MAMP right now. So i'm not able to test this very well. Uploading files works for me, but I can't check for image sizes.

Would you both mind checking over this? It looks like resizing from the main images uploaded (not in a field) wasn't working. But I can't be sure!

includes/class-assets.php
```diff
@@ -208,6 +210,24 @@ public function load_css() {
...
+	 * @since x.x
+	 */
+	public function image_listing_css() {
+		$thumbnail_width      = wpbdp_get_option( 'thumbnail-width' );
+		$thumbnail_height     = wpbdp_get_option( 'thumbnail-height' );
+		wp_add_inline_style(
+			'wpbdp-base-css',
+			'.wpbdp-listing .listing-thumbnail img {
+				max-width: ' . $thumbnail_width .'px;
+				max-height: ' . $thumbnail_height .'px;
```
stephywells, line 226: These need escaping. Also, we could use the 'Crop thumbnails to exact dimensions?' setting. If that is true, we'll need CSS cropping rather than just max sizes.

I think this would be cleaner if we keep it all together in the load_css() function. We should add vars to that rather than extra inline css.

stephywells (changes requested):

includes/class-assets.php
theRealRizeo: Thanks for the changes :) I have updated the css. Will continue on the tests on the images

theRealRizeo:
Done some tests @stephywells , crop works well checked and unchecked :)

garretlaxton:
Looks good still!

includes/class-assets.php
stephywells: Thanks Paul. There will be more to this than just the height and width. This would be best to add css vars, just like we did in that same function. Then we'll use them in the css. If cropping is happening, we need to use "object-fit: cover" to keep images proportionate. Please use your browser tools to inspect how we are doing this on the submit listing form.

theRealRizeo: Will do :)

theRealRizeo: Did some changes and added this CSS 
```
.wpbdp-thumbnail {
    max-width: var(--bd-thumbnail-width);
    max-height: var(--bd-thumbnail-height);
    &-cropped {
        width: var(--bd-thumbnail-width);
        height: var(--bd-thumbnail-height);
        object-fit: cover;
    }
}
```

theRealRizeo: With cropped enabled, the class `wpbdp-thumbnail-cropped` is added

assets/css/less/wpbdp.less
```diff
@@ -955,57 +955,16 @@ a.wpbdp-button {
...
-}
-
-.wpbdp-wp-theme-twentynineteen {
-
-    .wpbdp-listing .social-fields {
-        .twitter.twitter-handle {
-            vertical-align: text-top;
-        }
+.wpbdp-thumbnail {
+    max-width: var(--bd-thumbnail-width);
```
stephywells, line 959: Let's add a fallback in here for each of these lines so it'll look decent in older browsers too.
max-width: 200px;
max-width: var(--bd-thumbnail-width);

stephywells (changes requested):
This is looking great! Just one more change, and I think we can merge this,

Nice job on this!

assets/css/less/wpbdp.less
theRealRizeo: Done :)

theRealRizeo:
@garretlaxton there are some CSS adjustments done, please check if its still the same :)

garretlaxton (approved):
Images still look good!

stephywells (approved):
Yeah! This is all done and looks great!

PR #90, opened 2021-12-13
Divi builder compatibility

theRealRizeo:
This fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4937

Fixes support for Divi builder

### DIFF
--- a/includes/class-wordpress-template-integration.php
+++ b/includes/class-wordpress-template-integration.php
@@ -22,8 +22,10 @@ class WPBDP__WordPress_Template_Integration {
         }
 
         add_filter( 'template_include', array( $this, 'template_include' ), 20 );
-        add_action( 'wp_head', array( $this, 'maybe_spoof_post' ), 100 );
-        add_action( 'wp_head', array( $this, 'wp_head_done' ), 999 );
+		if ( ! WPBDP_Divi_Compat::divi_builder_is_active() ) {
+			add_action( 'wp_head', array( $this, 'maybe_spoof_post' ), 100 );
+			add_action( 'wp_head', array( $this, 'wp_head_done' ), 999 );
+		}
         add_filter( 'post_class', array( $this, 'post_class' ), 10, 3 );
     }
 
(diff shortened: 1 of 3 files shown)


garretlaxton:
@theRealRizeo Where can I find Divi? I don't see it in our team accounts.

theRealRizeo:
There are team accounts? I have it here https://drive.google.com/drive/folders/1ZdsSUgxCbCJOJ_yTXGBH1WsY2Ovzwbws?usp=sharing

garretlaxton:
@theRealRizeo By team accounts I was referring to the team password manager. Most things we use can be found there.

Relating to this PR, I'm currently not seeing any Divi options when editing a listing on the backend as seen in Salman's screenshot. I can see the Divi builder for regular WP posts, just not listings. Is there anything special I need to do?

theRealRizeo:
Hey @garretlaxton yes, you need to enable Listings under the Divi settings :) https://www.elegantthemes.com/blog/theme-releases/divi-custom-post-types

garretlaxton (approved):
Builder works good on my end!
Screenshot 2021-12-15 at 17-24-00 Edit Listing ‹ BD QA — WordPress

stephywells (commented):
Thanks for fixing this Paul!

But I'm confused here.... Do you know what maybe_spoof_post is doing? If it's not needed with Divi, does that mean we don't need it at all?

theRealRizeo:
Hey @stephywells I'll check what its doing. Maybe it was needed in older versions of WP. I will check what it does

theRealRizeo:
The spoofing helps to render the directory views. Without it, categories return 404 and listings have no titles. I think its also related to how the dispatcher works for the frontend views

stephywells:
@theRealRizeo does that mean this change will cause category 404s when Divi is installed?

theRealRizeo:
@stephywells the change does not affect normal usage of the categories or the pages (no 404 pages will be returned). Though this only fixes the backend editing using Divi. The frontend editing works, but not all parts of the listing are loaded

stephywells (changes requested):
It sounds like this is just about ready then. I've got just one suggestion.

stephywells (approved):
This looks great! Thanks so much Paul!

PR #95, opened 2021-12-20
Listing title single view as part

theRealRizeo:
Most themes that extend have the listing title section repeated. This will make it easier to include the same title template part in all themes


stephywells (approved):
Looks great. Thank you!

PR #96, opened 2021-12-20
Delete directory clear cache

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5008

This clears the cache when a page with the `[businessdirectory]` is permanently deleted from the database
If a page has a status of `trash` and has the shortcode `[businessdirectory]`, this is excluded. Trash pages tend to have the `/?page_id=xxx` in the url and could bring issues when an old directory page is moved to trash to make way for another directory page (please advise if this is okay)

### DIFF
--- a/includes/admin/class-admin.php
+++ b/includes/admin/class-admin.php
@@ -87,6 +87,9 @@ if ( ! class_exists( 'WPBDP_Admin' ) ) {
 			add_filter( 'admin_head-edit.php', array( $this, 'maybe_highlight_menu' ) );
 			add_filter( 'admin_head-edit-tags.php', array( $this, 'maybe_highlight_menu' ) );
 
+			add_filter( 'pre_trash_post', array( $this, 'before_trash_post' ), 10, 2 );
+			add_filter( 'pre_delete_post', array( $this, 'before_delete_post' ), 10, 3 );
+
 			require_once WPBDP_PATH . 'includes/controllers/class-addons.php';
 			WPBDP_Addons_Controller::load_hooks();
 
@@ -1215,6 +1218,31 @@ if ( ! class_exists( 'WPBDP_Admin' ) ) {
 
             echo '<script>jQuery(document).ready(function(){wpbdpSelectSubnav();});</script>';
         }
+
+		/**
+		 * Action called before a post is trashed.
+		 *
+		 * @since x.x
+		 */
+		public function before_trash_post( $check, $post ) {
+			$this->before_delete_post( $check, $post, false );
+			return $check;
+		}
+
+		/**
+		 * Action called before post is deleted.
+		 * Delete transient of saved directory ids if a directory page is deleted.
+		 *
+		 * @since x.x
+		 */
+		public function before_delete_post( $check, $post, $force_delete ) {
+			if ( 'page' === $post->post_type ) {
+				if ( strpos( $post->post_content, '[businessdirectory]' ) !== false ) {
+					wpbdp_delete_page_ids_cache();
+				}
+			}
+			return $check;
+		}
 
         /**
          * This function restores Manage Regions menu for Editors,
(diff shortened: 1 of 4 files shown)


garretlaxton (commented):
Trying to create the issue again I found that if the page is saved as `draft` or `pending` then this issue will still occur.

theRealRizeo:
Improved on this https://github.com/Strategy11/business-directory-plugin/pull/56 changed the query to allow only private and public pages  as they are the only pages with a full permalink

garretlaxton:
When the page is marked as `draft` or `pending`, the URLs seem to point to the next published page. 

**Example:**
In my case my page is `localhost/bd/`, but listing and category URLs point to the default WP sample page `localhost/sample-page/bd-listing/`

Here's a video which might be easier to understand: https://www.loom.com/share/96ec88ebbe7b4e05a25588763ab1bcd5

theRealRizeo:
@garretlaxton just pushed a fix for this. I have re-added the cache clear as I did not know if it was needed. The category links should match now

garretlaxton:
I can now access the listings, but my URLs still point to the default WP `sample-page`. I reset my permalinks and cleared my cache with no luck. 
d0bbd0a06c09483eb111be6152d808af

theRealRizeo:
If sample-page is the first page with the `[businessdirectory]` shortcode it will point to this. I will add post-nag to show what the default listing page is

theRealRizeo:
@garretlaxton please check now. The directory page should be labeled in the admin backend. This base url should match all front urls

garretlaxton:
I'm a bit confused as to why the default WP sample-page is showing as my directory page. This is the default page that comes with a fresh WP install, there isn't any shortcode on this page. 
Screenshot 2021-12-23 at 13-21-15 Pages ‹ BD QA — WordPress
Screenshot 2021-12-23 at 13-21-47 Edit Page ‹ BD QA — WordPress

theRealRizeo:
@garretlaxton if the "Sample Page" had the shortcode before, it will remain in the cache. The current change is for the main issue which is a page with the `[businessdirectory]` is deleted, it will set the page to the next page that has the shortcode. 
If a page is updated and the shortcode is removed, it will still keep that page in the cache.

Could you try the following:

- Add the shortcode to the `Sample Page`
- Delete the Sample Page (send to trash only)
- Check again the list of pages on what is highlighted (BD page will either be private or public)

I still think its a good idea to have the page as a setting that the user defines. Letting the plugin select based on content isn't the best way

garretlaxton:
Moving the sample page to the trash does help by pointing to a correct BD page, but my sample page never had any BD shortcodes as you can see in the previous screenshot it's still the default WP page. 

I agree a setting would be useful to select the main BD page, but something in this PR is causing the default WP sample page to display as the main BD page. It should point to my BD or Business Directory page by default since the sample page never had any shortcodes to cache.
Screenshot 2021-12-23 at 16-16-39 Pages ‹ BD QA — WordPress

I'm not sure if this would have to be fixed separately.

theRealRizeo:
The issue was with the cache, which is cleared when a page with the shortcode is deleted or moved to trash. 
Do you mind running the test again @garretlaxton , wit:

-  a BD page that first gets deleted
- 2 BD pages that have the shortcode, only one should show as the directory page
- Delete one of the BD pages and the other should take over

garretlaxton:
When I do a fresh install of BD and WP, the sample page is the main directory page before any business directory pages are created. 
Screenshot 2022-01-02 at 23-14-26 Pages ‹ wordpress — WordPress

This prevents our plugin from showing the notification alerting the user to create a BD page. 
Screenshot 2022-01-02 at 23-12-22 Pages ‹ wordpress — WordPress

This only happens in this branch, so it seems something is causing the main directory page to point to the sample page.

theRealRizeo:
Hey @garretlaxton thanks for the additional tests :) I have made some changes to the SQL only, to check for pages that are either private or published. The cache will also be cleared for a fresh lookup to be checked and it will always get the first published page with the shortcode

garretlaxton:
Looks great now! I just notice the `Directory Page` banner is missing.
Screenshot 2022-01-04 at 14-43-31 Pages ‹ wordpress — WordPress

theRealRizeo:
Yeah, removed it for now unless its a feature we really want :)

garretlaxton (approved):
Gotcha!

stephywells (approved):
I went through and cleaned this up. It looks like this cache wasn't actually doing anything other than causing extra queries. This is updated now to use our caching system for all db calls.

PR #97, opened 2021-12-22
Handle missing templates silently

theRealRizeo:
Removes exception thrown in case a template file does not exist.
If the file does not exist, return an empty string

### DIFF
--- a/includes/themes.php
+++ b/includes/themes.php
@@ -613,7 +613,8 @@ class WPBDP_Themes {
 
     private function render_template_file( $template_id, $path, $vars = array() ) {
         if ( ! $path ) {
-            throw new Exception( 'Invalid template path for template: "' . $template_id . '"' );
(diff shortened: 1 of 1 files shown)


stephywells:
Was this causing a problem? Do you have more info on what was happening before?

theRealRizeo:
Before if there was no file, it would show the WSOD and send the admin the error email to recover the site

theRealRizeo:
I tested by changing or adding 1 word on the included template file, before the patch

stephywells:
Was there a template that is missing? Or were you only seeing this during development?

theRealRizeo:
Its related to this PR https://github.com/Strategy11/bd-business-card/pull/2#discussion_r773514879
It would be useful if an update is done and the template file does not exist

stephywells:
Gotcha. Thanks for clearing that up! The PR looks good.

stephywells (approved):

PR #105, opened 2022-01-10
Issue 5012

theRealRizeo:
Linked issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5012

Fixes the following issues:

- New module installed : New modules installed and activated would deactivate all licenses
- License deactivation : License checks would cause deactivation

I separated the check for update data that calls `get_version` to a separate function, to be able to later on change this easily to the new rest endpoint.


garretlaxton:
I'm going to keep watching this a bit more. I have 1 site with the patch and another without. I'm waiting for the unpatched to deactivate so I can see if the patched remains activated.

The more I test this the less I know how to actually reproduce this haha.

stephywells:
This is super hard to reproduce consistently. It looks like each time a new add-on is activated, the cache on the site is incomplete, so it triggers an API request. When the license key is missing from the response, it gets deactivated.

I'm not finding a way that deactivates all license keys though. I'm only able to get one deactivated. From what I'm seeing, it doesn't look like this change makes any difference. I'm not seeing any plugins or themes that have a mismatch for the name.

Here's what helps for me:
Screen Shot 2022-01-11 at 3 20 46 PM

theRealRizeo:
On my end, when comparing names, the license key kept deactivating. Steps to reproduce:

- Use an addon that has never been activated on the site with the Free version. If it has been activated before on the free version it will not work
- Once activated, it should only ask for the license of that one module, and not the other activated modules.

If the addon has been activated before on the lite version, the error will not be possible to reproduce

garretlaxton:
I haven't had my license deactivate on my site with this patch so far.

stephywells:
@theRealRizeo I'm just trying to replicate the issue with the name not matching. I completely agree that it would be good to change, but I'd like to understand what you saw. Do you have an example of names that weren't matching for you?

stephywells:
Thanks for checking that @theRealRizeo. In that case, I'm not seeing any changes here that would affect licenses getting deactivated. Can you explain a bit more about the problem you saw, and how this fixes it?

includes/licensing.php
```diff
@@ -999,20 +991,19 @@ public function get_version_information( $force_refresh = false ) {
         }
 
         $updates = array();
-
         foreach ( $this->items as $item ) {
             $item_key = $item['item_type'] . '-' . $item['id'];
 
             foreach ( $body as $item_information ) {
-				if ( $item_information->name !== $item['name'] ) {
+				if ( $item_information->slug !== $item['id'] ) {
```
stephywells, line 998: I'm looking into this a bit more, and it looks like this prevents modules from getting updates since the id and slug don't match. Here are a few examples:
slug: migrator, id: business-directory-migrate
slug: enhanced-categories, id: business-directory-categories
slug: elegant-grid-theme, id: elegant-grid

Is this tested for each module and theme?

stephywells: Maybe we should try testing this one?
```
if ( $item_information->name !== $item['name'] || empty( $item_information->license ) ) {
continue;
}
```

garretlaxton: Can confirm this is blocking my site from receiving the ratings & business card theme updates.

theRealRizeo: Did some tests with the master branch and I also do not get plugin update notices. Was there an API change that could've affected this?

stephywells: No, there wasn't an api update for this. It just depends on the product. Some match the slug and some don't.

theRealRizeo: Did more tests, I changed it to :

```
if ( trim( $item_information->name ) !== trim( $item['name'] ) || empty( $item_information->license ) ) {
      continue;
}
```

For update checks, these have to manually be triggered at the updates screen:

image

I did not see issues with other activated modules being deactivated also

stephywells:
@theRealRizeo what's the status on this?

theRealRizeo:
@stephywells set it as ready for QA. Already pushed changes for it. @garretlaxton please test :)

garretlaxton (approved):
Looks good! Haven't had to reactivate my license.

PR #106, opened 2022-01-11
Fix Field value filter

theRealRizeo:
Fixes missing parameter in the `wpbdp_form_field_html_value` in the text area field type.
The filter requires 4 parameters but only 3 were present. This mainly affects using the filter in the Field Icons premium PR where the filter defined in https://github.com/Strategy11/business-directory-plugin/blob/816824cf4d96a4c53de71106ad73c54eda058f2e/includes/fields/class-form-field.php#L435 has 4 parameters

### DIFF
--- a/includes/fields/class-fieldtypes-textarea.php
+++ b/includes/fields/class-fieldtypes-textarea.php
@@ -317,7 +317,7 @@ class WPBDP_FieldTypes_TextArea extends WPBDP_Form_Field_Type {
     }
 
     public function get_field_html_value( &$field, $post_id ) {
-        $value = apply_filters( 'wpbdp_form_field_html_value', $field->value( $post_id ), $post_id, $field );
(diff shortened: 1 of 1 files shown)


stephywells:
I don't think we can set this for sure. It doesn't look like this would be accurate when it's an excerpt.

But I don't think we need it. If we need four parameters, it can be done like the existing maybe_shorten_output_in_excerpt function. But I'm not seeing this fourth param used in the function in Premium anyway...

I think we can go ahead and close this one without merging. I think it's more likely to cause issues than the solve them.

theRealRizeo:
Regarding the PR for field icons, its is in use, and for PHP version 7.3 and above, this throws a fatal error

theRealRizeo:
I will check on how best this can be improved

theRealRizeo:
I will check on how best this can be improved

theRealRizeo:
Okay, removing the third parameter wont affect how the filter works in the Field icons PR. Closing

PR #108, opened 2022-01-12
Fee Plan Table Fix

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5015

This fixes the `Listings` column in the `Fee Plans` Table. Reported issues were :

- Admin submitted paid listings would count towards the total. This would give a false total count of the revenue the specific plan collected
- A listing created with a 100% discount would show the total as `free` and not `0.00`. This would only happen if there is one listing of such type with the paid plan. If there are more than one listings, the total count shows correctly.

image

Expected view after changes

image

### DIFF
--- a/includes/admin/helpers/tables/class-fees-table.php
+++ b/includes/admin/helpers/tables/class-fees-table.php
@@ -254,7 +254,7 @@ class WPBDP__Admin__Fees_Table extends WP_List_Table {
 			return $column;
 		}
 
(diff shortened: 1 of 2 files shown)


garretlaxton (approved):
Looks good! Tested with a few other discount increments besides 100% off.

stephywells (approved):
This looks great! Thanks so much Paul!

PR #109, opened 2022-01-12
Field Wrapper adjustments

theRealRizeo:
https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5016

Add extra class filter `wpbdp_display_field_wrapper_classes` to allow adding of extra classes to the field wrapper.

The filter is being used in this line https://github.com/Strategy11/business-directory-premium/blob/6038eab51e7927e4de33136a3ffd8a495315e51a/classes/class-wpbdp-field-icon.php#L23 to include an extra class in the wrapper. 
Added an extra class to hide the label if the value is to be used as an icon only
End result is as follows 

image


stephywells (approved):
This looks good to me. Thanks Paul!

PR #113, opened 2022-01-14
Field Label

theRealRizeo:
Remove Field label colon and add padding on the right


stephywells (approved):
Looks great. Thanks @theRealRizeo !

PR #116, opened 2022-01-18
Fix file upload for field images

theRealRizeo:
Fix issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5019

Fixes an issue where custom social icons were not being uploaded.
Image file types were not being uploaded from the frontend due to a sanitization that handling the file types as a text field.


garretlaxton (commented):
Would there be a better way to sanitize this instead of removing it?

theRealRizeo:
Not that I have seen. File uploads cannot be sanitized, but checked that the content is valid and the size does not exceed a certain limit.

stephywells (changes requested):
It looks like we need to follow this through and add sanitizing elsewhere.

For example, it looks like we should replace line 279 ($files,) with 'file',

By doing it this way, we will end up with a lot fewer spots that use the $_FILES directly.

Then, in the utils media_upload() function, add:
if ( $file_ && is_string( $file_ ) && strpos( $file_, '/' ) === false ) { // This should be a better check
  $file_ = wp_unslash( $_FILES[ $file_ ] );
}

Now, we'll need sanitizing. It looks like it might make the most sense to add on line 265:
$file = $file_;

Something like:
$file = $file_;
$file['error'] = sanitize_text_field( $file['error'] );

We can sanitize each part of the $file array that we will be using. This will need testing and a deeper look into the files array to see what cannot be sanitized. But let's not remove all the sanitizing.

theRealRizeo:
The `$file['error']` expects an integer, using `$file['error'] = sanitize_text_field( $file['error'] );` at that point converts the integer to a string and fails the file upload. 
There is a check in the function `wp_handle_upload` and `wp_handle_sideload` that checks if the error is an integer of 0 before proceeding the upload. 
I moved the `wp_unslash` to be in the function. 

The `$_FILES` object contains:

```
(
	[name] => MyFile.txt (comes from the browser, so treat as tainted)
	[type] => text/plain  (not sure where it gets this from - assume the browser, so treat as tainted)
	[tmp_name] => /tmp/php/php1h4j1o (could be anywhere on your system, depending on your config settings, but the user has no control, so this isn't tainted)
	[error] => UPLOAD_ERR_OK  (= 0)
	[size] => 123   (the size in bytes)
)
```

There are already checks done within the WordPress functions that further validates uploads into the media library.
The `$_FILES` variable used within the `media_upload` function in `utils.php` class is used to get or set (for single uploads) the file id from the function [media_handle_upload](https://developer.wordpress.org/reference/functions/media_handle_upload/)

theRealRizeo:
I don't think any further sanitization can be done on the `$_FILES` object that will not affect file uploads

stephywells:
@theRealRizeo I think moving that unslash could potentially cause a lot of issues. Let's not do it this way, unless we are using it directly on a $_FILES array.

Also, it looks like our get_mimetype() could definitely use some sanitizing.

garretlaxton (approved):
Looks good here.

includes/utils.php
```diff
@@ -312,20 +312,16 @@ public static function media_upload( $file_, $use_media_library = true, $check_i
...
+            $mime_type = mime_content_type( $file );
 		}
 
 		if ( null === $mime_type ) {
 			$type_info = wp_check_filetype( $file, wp_get_mime_types() );
 			$mime_type = $type_info['type'];
 		}
 
-		return $mime_type;
+		return esc_attr( $mime_type );
```
stephywells, line 324: We should be sanitizing here instead of escaping since this in incoming data rather than echoing.

includes/utils.php
```diff
@@ -312,20 +312,16 @@ public static function media_upload( $file_, $use_media_library = true, $check_i
...
 		$mime_type = null;
 
-		if ( function_exists( 'finfo_open' ) ) {
-			if ( $finfo = finfo_open( FILEINFO_MIME ) ) {
-				$mime_type = explode( ';', finfo_file( $finfo, $file ) );
-				$mime_type = trim( $mime_type[0] );
-				finfo_close( $finfo );
-			}
+		if ( function_exists( 'mime_content_type' ) ) {
+            $mime_type = mime_content_type( $file );
```
stephywells, line 316: Can you please fix the tabs in this line?

stephywells (changes requested):

includes/utils.php
theRealRizeo: Thanks, I have added sanitization :)

stephywells (approved):
Looks good!

PR #118, opened 2022-01-26
Cache Issue

theRealRizeo:
Related issue : https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5022

The current implementation to use the `WP_Cache` object sets the cache, but it gets cleared on every new request. I implemented some tests on this and each result was `false` on the cache. [WordPress Cache tests](https://github.com/WordPress/wordpress-develop/blob/trunk/tests/phpunit/tests/cache.php) run on a local variable emulating the cache system. In this type of test case, the value will be persisted at each point of the tests as the object is loaded once. In the plugin case, it seems the call is done about 10 times at different parts of the page loading time.

EDIT : 

Added a transient cache layer to save a cache of the data for the specified timeframe. I did not notice duplicate queries related to the plugin when navigating the plugin frontend.
The cache saved the cache key value in a group as an key-value array `$data = array( 'key' => 'value')` The entire array is saved in the transient to the respective group `set_transient( $group, $data , $time );` 
I found it easier to add the function to the existing cache functions to handle other areas of the plugin.

EDIT 👍🏾 
Cache to only fix getting the plugin page ids


garretlaxton:
Something I'm seeing on query monitor is the following query duplicated 4 times when viewing a listing.
Screenshot 2022-01-27 at 05-38-18 Andy Agent

theRealRizeo:
Thanks, let me check all the queries :)

theRealRizeo:
Closing for now as I change the code

theRealRizeo:
@garretlaxton Did some changes to add an extra cache layer. Regarding `wpbdp_get_post_by_id_or_slug` , it only does a single Id call and can be permitted at this point. Adding a cache of listing ids might be slower than performing an actually query

stephywells:
We definitely don't want to be adding transients for every single query like this. The point is to add integration with server caching, but not totally replace it.

Let's address this issue directly, rather than changing the way the whole cache works. This will add far too much load.

stephywells (changes requested):
The idea of this looks good, but the problem is we can't do this universally because it could really bog down the database. We need this issue resolved right away, but only make changes in the wpbdp_get_page_ids_with_query() and wpbdp_delete_page_ids_cache() functions.

theRealRizeo:
@garretlaxton updates done:

- Removed global change on cache function, the database cache is only set for checking the BD pages (the reported duplicate query)
- A transient will be used to save the cache data incase object cache is not available
- If object cache is available via Redis, Memcached or another system, this will be used and the transient will not be used

includes/utils.php
```diff
@@ -231,6 +231,54 @@ public static function cache_delete_group( $group ) {
...
+	 * @param string $args[group] The name of the cache group
+	 *
+	 * @since x.x
+	 *
+	 * @return mixed
+	 */
+	public static function check_transient_cache( $args ) {
+		$results = get_transient( $args['group'] );
+		$key     = $args['cache_key'];
+		if ( false !== $results && is_array( $results ) && isset( $results[ $key ] ) ) {
```
stephywells, line 248: Let's simplify this down a bit:
if ( is_array( $results ) && isset( $results[ $key ] ) ) {

includes/utils.php
```diff
@@ -231,6 +231,54 @@ public static function cache_delete_group( $group ) {
...
+			'time'      => 300,
+		);
+		$args = array_merge( $defaults, $args );
+
+		$cache = get_transient( $args['group'] );
+		if ( ! $cache || ! is_array( $cache ) ) {
+			$cache = array();
+		}
+		$cache[ $args['cache_key'] ] = $args['results'];
+		set_transient( $args['group'], $cache, $args['time'] );
```
stephywells, line 279: If one value in the group gets updated frequently, and other don't, this could hold expired values for other items in the group. It might be better to use an option, and include the expirations for each value in the array.

includes/helpers/functions/general.php
```diff
@@ -93,14 +93,38 @@ function wpbdp_get_page_ids_with_query( $page_id ) {
 
 	$q .= ' ORDER BY ID DESC ';
 
-	return WPBDP_Utils::check_cache(
+	if ( ! wp_using_ext_object_cache() ) {
+		$results = WPBDP_Utils::check_transient_cache(
+			array(
+				'cache_key' => 'wpbdp_page_ids' . $page_id,
+				'group'     => 'wpbdp_pages',
```
stephywells, line 100: I'm thinking we should either simplify this a lot, or send 'query' and 'type' in this array so it can all be handled in one place.

The simple way:
```
if ( ! wp_using_ext_object_cache() ) {
$results = get_transient( 'wpbdp_page_ids' . $page_id );
if ( $results !== false ) {
  return $results;
}
}

$results = WPBDP_Utils::check_cache( ... );
if ( ! wp_using_ext_object_cache() ) {
set_transient( 'wpbdp_page_ids' . $page_id,  $results );
}
```

stephywells (changes requested):
Thanks for simplifying this @theRealRizeo! I'm still seeing room for bugs that we can root out before merging.

includes/helpers/functions/general.php
theRealRizeo: I think having it all as an array in an option would be easier as it makes it easier to clear the cache

theRealRizeo:
@stephywells made the changes as suggested:

- Cache array saved as a multidimensional array in an option that does not autoload
- The inner array contains the option value and the expire time in miliseconds
- When getting the cache value from the options, the expire time is validated. If the cache time is past the set time, it is removed and the option is updated.

includes/helpers/functions/general.php
stephywells: If we use an option, I think we do want it to autoload since this is a query that runs on every page. This option solution still has the issue of possibly never refreshing on option that isn't called often. If we do it this way, the expiration needs to be saved for each option individually.

There are only 4 possible $page_id parameters here, so clearing the cache isn't complicated.
https://github.com/Strategy11/business-directory-plugin/blob/78e6ac7a1a9bcf2af1913b90e6f711b54d7cfded/includes/helpers/functions/general.php#L25-L29

So, we still need to either do this the simple way, or handle these three functions from the WPBDP_Utils class, by sending 'query' and 'type' to the array.

With all the complication this adds, I would vote the simple option that will be much less likely to introduce bugs.

stephywells (changes requested):

stephywells:
@garretlaxton does this work for you now?

garretlaxton:
@stephywells It works yes, but I am still seeing the duplicate queries on a listing relating to `wpbdp_get_post_by_id_or_slug()` which isn't something I see on my demo without the update.

stephywells:
Thanks @garretlaxton ! Looks like that's not related to this change. It looks like the site you're comparing to has the advanced CPT turned on. I'll go ahead and merge this, then add the cache to the wpbdp_get_post_by_id_or_slug() function too.

PR #119, opened 2022-01-28
Import file extension convert to lowercase

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5024

Currently if an import is uploaded with the `.csv` in uppercase as `.CSV` the error returns that it is an invalid format. This fix sets the extension to lower case to avoid case-sensitive validation of files uploaded

### DIFF
--- a/includes/admin/csv-import.php
+++ b/includes/admin/csv-import.php
@@ -440,7 +440,7 @@ class WPBDP_CSVImportAdmin {
 			'csv'    => 'csv',
 		);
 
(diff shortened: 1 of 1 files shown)


garretlaxton (approved):
Looks good! Extensions are no longer case sensitive, both CSV and ZIP.

stephywells (approved):
Perfect! Thanks Paul!

PR #120, opened 2022-01-28
Proper handling of removing images

theRealRizeo:
Issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5014

There was an issue where different listings sharing images would not delete the images. This would mainly happen if the first shared image is the same shares the same attachment id.
During the fix process, I noticed the attachment cache was not cleared, and remained with the listing, also the re-arrange function in the frontend would get an attachment object instead of a listing object. This resulted in a broken image display of a non-existent image. 

Using the function `clean_post_cache` thhe cache is cleared and there is also an extra check if the selected image is the thumbnail to also exclude it and get a fresh image from the list.


garretlaxton (approved):
Looks good, images with the same ID in multiple listings delete as should.

stephywells (approved):
Nice job on this @theRealRizeo ! Great job!

PR #121, opened 2022-02-01
Fix Issue 5021 - Submit listing with one category

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5021

This also fixes the issue with the category step showing up in the steps when moving to step 2 in this scenario.


garretlaxton (approved):
Looks good! Also retested with payments on.
Screenshot 2022-02-01 at 17-10-41 Add Listing »
Screenshot 2022-02-01 at 16-59-32 Add Listing »

PR #124, opened 2022-02-02
Issue 5029

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5029

A new admin listing will need a plan assigned to it. This introduces an admin ajax function to assign the plan to the listing in the backend. If a plan has no listing, a message, `Please select a plan before uploading images to the listing`, will be shown to the user to first select a plan before uploading images


garretlaxton (approved):
Looks good.

No Plan:
Screenshot 2022-02-03 at 18-48-21 Add New Listing ‹ BD QA — WordPress

Plan Assigned:
Screenshot 2022-02-03 at 18-48-48 Add New Listing ‹ BD QA — WordPress

includes/admin/controllers/class-admin-listings.php
```diff
@@ -700,6 +702,32 @@ public function ajax_clear_payment_history() {
...
     }
 
+	/**
+	 * Assign a plan to a listing.
+	 * This assigns a plan to a listing in the admin backend.
+	 *
+	 * @since x.x
+	 */
+	public function ajax_assign_plan_to_listing() {
+		check_ajax_referer( 'wpbdp_ajax', 'nonce' );
```
stephywells, line 712: Looks like we're missing the permission check here.

includes/admin/controllers/class-admin-listings.php
```diff
@@ -700,6 +702,32 @@ public function ajax_clear_payment_history() {
...
+			wp_send_json_error();
+		}
+		$listing = wpbdp_get_listing( $listing_id );
+		$current_plan = $listing->get_fee_plan();
+
+		if ( ! $current_plan || (int) $current_plan->fee_id != (int) $plan_id ) {
+			$payment = $listing->set_fee_plan_with_payment( $plan_id );
+
+			if ( $payment ) {
+				$payment->process_as_admin();
```
stephywells, line 725: Could this add a free payment to the listing or mark a pending listing as complete? Can you help me understand why the payment needs to be included here?

stephywells (changes requested):

includes/admin/controllers/class-admin-listings.php
theRealRizeo: Permission check added to match metabox check

includes/admin/controllers/class-admin-listings.php
theRealRizeo: Agreed, this would submit the payment which we do not want and can cause multiple payments saved with each plan change. I've removed the line after `$listing->set_fee_plan_with_payment( $plan_id );`

includes/admin/controllers/class-admin-listings.php
```diff
@@ -700,6 +702,31 @@ public function ajax_clear_payment_history() {
...
+		check_ajax_referer( 'wpbdp_ajax', 'nonce' );
+		$listing_id = wpbdp_get_var( array( 'param' => 'listing_id' ), 'post' );
+		$plan_id    = wpbdp_get_var( array( 'param' => 'plan_id' ), 'post' );
+		if ( ! $listing_id || ! $plan_id ) {
+			wp_send_json_error();
+		}
+		$listing = wpbdp_get_listing( $listing_id );
+		$current_plan = $listing->get_fee_plan();
+
+		if ( ! $current_plan || (int) $current_plan->fee_id != (int) $plan_id ) {
```
stephywells, line 724: Let's use !== whenever possible.

stephywells (changes requested):

includes/admin/controllers/class-admin-listings.php
stephywells: It looks like that function creates a payment too. Is that needed?

theRealRizeo: Changed it to use `set_fee_plan` which doesn't create a payment. Payment will only be created on submit but not on change

stephywells: Looks good!

stephywells:
@theRealRizeo do we need one more round of testing on this from Garret?

stephywells (approved):
This looks good now. Thanks Paul!

theRealRizeo:
@garretlaxton please check again, some code changes have been implemented

garretlaxton:
I still see the error before I select a plan, but after selecting a plan it shows the original error and I can't upload an image.
Screenshot 2022-02-24 at 18-05-41 Add New Listing ‹ BD QA — WordPress


Also, the **Select Media** button still allows you to upload as many images as you want before choosing a plan. Not sure if we want that consistent with how the dropzone acts?
Screenshot 2022-02-24 at 17-57-31 Add New Listing ‹ BD QA — WordPress

theRealRizeo:
Issue seemed to be when selecting a plan and changing it. I added an option of "None" in the drop down to allow select change in the drop down

Untitled

This allows for selecting a plan to reflect in the backend and attempt to fix the json parse error on select

theRealRizeo:
@garretlaxton  I've simplified the JS as I noticed at most time the `parseJSON` would fail and the image selection would not work well. Regarding the uploader, there is no limit defined in the frontend, but the validation is handled in the backend

garretlaxton (approved):
Functions great now!

includes/admin/controllers/class-admin-listings.php
```diff
@@ -700,6 +702,41 @@ public function ajax_clear_payment_history() {
...
+		if ( ! $listing_id || ! $plan_id ) {
+			wp_send_json_error();
+		}
+		$listing = wpbdp_get_listing( $listing_id );
+		$current_plan = $listing->get_fee_plan();
+
+		if ( ! $current_plan || (int) $current_plan->fee_id !== (int) $plan_id ) {
+			$listing->set_fee_plan( $plan_id );
+		}
+        $new_plan = $listing->get_fee_plan();
```
stephywells, line 727: Pleas check tabs here

assets/js/admin-listing-metabox.js
```diff
@@ -87,12 +90,29 @@ jQuery( function( $ ) {
...
         } else {
             // Plan changes are handled in a special way.
-            if ($input.is('#wpbdp-listing-plan-select')) {
-                var plan = $.parseJSON( $input.find( 'option:selected' ).attr( 'data-plan-info' ) );
-
-                $metabox_tab.find('input[name="listing_plan[fee_id]"]').val(plan.id);
-                $metabox_tab.find('input[name="listing_plan[expiration_date]"]').val(plan.expiration_date);
-                $metabox_tab.find('input[name="listing_plan[fee_images]"]').val(plan.images);
+            if ( $input.is('#wpbdp-listing-plan-select') && $input.val() ) {
+                var listing_id = $( 'input[name="post_ID"]' ).val();
```
stephywells, line 94: Please check tabs on changes in this file too

stephywells (changes requested):
This is almost there.

stephywells (approved):
Thanks for the fix and testing! Looks good.

PR #126, opened 2022-02-03
Allow enable and disable of default plan

theRealRizeo:
Related to https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5031

Current functionality:
- Default plan only gets enabled if there is no payment gateway active. 
- If there is a payment active, it is set to disabled to only allow paid plans to show in the frontend

New functionality:
- Default plan can be enabled or disabled regardless of payment plan state
- If there is payment gateway enabled, the default plan will not be visible in the frontend
- If there is no payment gateway available and the default plan is active, it will show in the frontend
- If there is no payment gateway available and the default plan is disabled, it will not show in the frontend


stephywells (approved):
Awesome!

PR #127, opened 2022-02-03
Remove order sorting when its not custom order

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5030

Show the order column only if the "Custom Order" section is selected

image

image


stephywells (approved):
Looks good to me! Thank you Paul!

PR #131, opened 2022-02-11
Only delete image if not attached to another post

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5034

Before images linked to other posts or pages would be deleted which is something we do not want. 
This fix checks if the attachment is linked to a post via the post meta name `_thumbnail_id` . If there is another post linked to the image, it will not be deleted. It is not currently possible to check for images added in a post body as they are not recorded in the post_meta table.

### DIFF
--- a/includes/helpers/class-listing-image.php
+++ b/includes/helpers/class-listing-image.php
@@ -125,8 +125,9 @@ final class WPBDP_Listing_Image {
 
 		// If the current listing is still cached, don't count it.
 		$linked_listings = array_diff( $linked_listings, array( $listing_id ) );
-
-		if ( empty( $linked_listings ) ) {
+		$linked_posts    = self::get_linked_posts( $id );
+		$linked_posts    = array_diff( $linked_posts, array( $listing_id ) );
+		if ( empty( $linked_listings ) && empty( $linked_posts ) ) {
 			wp_delete_attachment( $id, true );
 		} else {
 			// Attach to the next listing.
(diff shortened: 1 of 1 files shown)


includes/helpers/class-listing-image.php
```diff
@@ -125,8 +125,9 @@ public static function maybe_delete_image( $id, $listing_id ) {
 
 		// If the current listing is still cached, don't count it.
 		$linked_listings = array_diff( $linked_listings, array( $listing_id ) );
-
-		if ( empty( $linked_listings ) ) {
+		$linked_posts    = self::get_linked_posts( $id );
```
stephywells, line 128: Let's only check for linked posts if there are no linked listings to cut down on DB calls. Also, it would be great to move some of this to a new function that returns true or false if there is another post or listing using the image.

includes/helpers/class-listing-image.php
```diff
@@ -138,4 +139,26 @@ public static function maybe_delete_image( $id, $listing_id ) {
...
+	 * Checks id an attachment is linked to other posts.
+	 * This checks if the attachment is in use with other posts and returns the ids.
+	 *
+	 * @param int $attachment_id The attachment id.
+	 *
+	 * @since x.x
+	 *
+	 * @return array
+	 */
+	public static function get_linked_posts( $attachment_id ) {
```
stephywells, line 153: Let's make new functions private whenever possible to make future maintenance easier.

stephywells (changes requested):

includes/helpers/class-listing-image.php
theRealRizeo: Changes applied. I have separated the check for listings and posts to a separate private function `check_if_attachment_is_linked` that returns a boolean and this is called once in the function `maybe_delete_image`

stephywells (approved):
Looks good! Thank you!!

PR #133, opened 2022-02-14
Edit Listing Fee Plan Error Message

theRealRizeo:
Reported issue : https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5033

This is a special case that should happen in a very rare occasion. Steps to reproduce

- Create a listing that has a premium plan assigned to it and publish it
- Edit the Fee plans and disable all plans and gateways
- Proceed to edit listing and an error message will show. Previous message was : `"Please choose a valid category for your plan."` . New message is : `"Current active fee plan is disabled. Please select another fee plan."` (may need better wording)

A better option would be to show a message on load or disable the "Next" button for a listing that fits the above criteria

image

image

### DIFF
--- a/includes/admin/controllers/class-admin-fees.php
+++ b/includes/admin/controllers/class-admin-fees.php
@@ -112,11 +112,11 @@ class WPBDP__Admin__Fees extends WPBDP__Admin__Controller {
 
 		if ( ! is_wp_error( $result ) ) {
 			if ( 'insert' === $mode ) {
-				wpbdp_admin_message( __( 'Fee plan added.', 'business-directory-plugin' ) );
+				wpbdp_admin_message( __( 'Plan added.', 'business-directory-plugin' ) );
 			} elseif ( $images_changed ) {
 				$this->show_update_listing_msg( $fee );
 			} else {
-				wpbdp_admin_message( __( 'Fee plan updated.', 'business-directory-plugin' ) );
+				wpbdp_admin_message( __( 'Plan updated.', 'business-directory-plugin' ) );
 			}
 		} else {
 			foreach ( $result->get_error_messages() as $msg ) {
@@ -131,7 +131,7 @@ class WPBDP__Admin__Fees extends WPBDP__Admin__Controller {
 	 * @since 5.15.3
 	 */
 	private function show_update_listing_msg( $fee ) {
-		$message = __( 'Fee plan updated.', 'business-directory-plugin' );
+		$message = __( 'Plan updated.', 'business-directory-plugin' );
 
 		$total_listings = $fee->count_listings();
 		if ( ! $total_listings ) {
@@ -173,11 +173,11 @@ class WPBDP__Admin__Fees extends WPBDP__Admin__Controller {
 		$fee     = wpbdp_get_fee_plan( $plan_id );
 		$res     = new WPBDP_Ajax_Response();
 		if ( ! $fee ) {
-			$res->send_error( __( 'Fee plan not found.', 'business-directory-plugin' ) );
+			$res->send_error( __( 'Plan not found.', 'business-directory-plugin' ) );
 		}
 
 		$this->update_listing_images( $fee );
-		$res->set_message( __( 'Fee plan listings updated.', 'business-directory-plugin' ) );
+		$res->set_message( __( 'Plan listings updated.', 'business-directory-plugin' ) );
 		$res->send();
 	}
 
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
Looks good, but is says select another plan which you can't do when editing the listing.
Screenshot 2022-02-14 at 15-23-40 BD QA – Just another WordPress site

theRealRizeo:
Yes, we may need to either disable the "Next" button when there are no enabled plans, which I think would be a good option for this

stephywells:
So it sounds like we need a different message when there are no plans to choose from. Was this an error with the migration that disabled a plan that shouldn't have been disabled?

Before a plan is disabled, can we check if there are other plans enabled? If not, we should show a message and don't disable the plan.

theRealRizeo:
@stephywells before the PR to remove the payments settings, the default fee plan could not be enabled or disabled, but would show on certain conditions. If payments were turned on, the default plan would not be shown and vice versa. 

> Before a plan is disabled, can we check if there are other plans enabled? If not, we should show a message and don't disable the plan.

Will work on this :)

stephywells:
But is it possible the migration disabled this plan on their site?

garretlaxton:
@stephywells Not sure what you are referring to by migration, but it was the 5.17 update I'm pretty sure. A bug in 5.17 disabled free plans when we removed the _turn payments on_ option.

theRealRizeo:
@stephywells before the default free plan would not be allowed to be disabled or enabled. When we allowed this option in this PR https://github.com/Strategy11/business-directory-plugin/pull/126 then is allowed for there to be a state of no plan active. Before the default plan would disable if there are payment gateways enabled and vice versa. 
Allowing the admin to enable/disable the default free plan would bring about this.

The migration did not affect as it only checked for the payment option, this scenario can be replicated on a fresh install by manually disabling all plans and trying to edit an existing listing that has a plan. Suggested options :

- We add a check before disabling a plan to alert the admin this will affect listings that will be edited
- Before all plans are disabled, we show a notice to the admin that they are about to disable creating a new listing in the frontend

includes/controllers/pages/class-submit-listing.php
```diff
@@ -890,7 +890,11 @@ private function plan_selection() {
 
             if ( ! $plan || ! $plan->enabled || ! $plan->supports_category_selection( $categories ) ) {
                 if ( $this->editing ) {
-                    $this->messages( _x( 'Please choose a valid category for your plan.', 'submit listing', 'business-directory-plugin' ), 'error', 'plan_selection' );
+					if ( ! $plan->enabled ) {
+						$this->messages( _x( 'Current active fee plan is disabled. Please select another fee plan.', 'submit listing', 'business-directory-plugin' ), 'error', 'plan_selection' );
```
stephywells, line 894: Let's use "plan" going forward instead of "fee plan". It's confusing since they don't have to have a fee.

includes/class-fees-api.php
```diff
@@ -68,6 +68,29 @@ public static function has_paid_plans() {
...
+
+    /**
+	 * Check if there are enabled plans.
+	 * This does a count for all enabled plans regardless of the amount.
+	 *
+	 * @since x.x
+	 *
+	 * @return bool
+	 */
+	public static function has_enabled_plans() {
```
stephywells, line 80: Can we reduce duplication here a bit by adding has_enabled_plans( $paid = null )

Then:
$cache_key = 'enabled_plan_count';
$query = "SELECT count(*) FROM {$wpdb->prefix}wpbdp_plans WHERE enabled != 0";
if ( $paid ) {
$cache_key = 'paid_plan_count';
$query .= "AND amount > 0";
}

includes/admin/controllers/class-admin-fees.php
```diff
@@ -251,9 +251,14 @@ function delete_fee() {
...
     function toggle_fee() {
 		$fee = $this->get_or_die();
-        $fee->enabled = ! $fee->enabled;
-        $fee->save();
-		wpbdp_admin_message( $fee->enabled ? _x( 'Fee enabled.', 'fees admin', 'business-directory-plugin' ) : _x( 'Fee disabled.', 'fees admin', 'business-directory-plugin' ) );
+		$enabled_plans = WPBDP_Fees_API::has_enabled_plans();
+		if ( $enabled_plans > 1 || ! $fee->enabled ) {
+			$fee->enabled = ! $fee->enabled;
+			$fee->save();
+			wpbdp_admin_message( $fee->enabled ? _x( 'Fee enabled.', 'fees admin', 'business-directory-plugin' ) : _x( 'Fee disabled.', 'fees admin', 'business-directory-plugin' ) );
```
stephywells, line 258: Let's use "plan" here too instead of "fee"

stephywells (changes requested):
Nice job! just a couple wording and code types of requests.

stephywells (approved):
Nice! This looks great!

PR #134, opened 2022-02-17
Fix duplicate titles on excerpt view

theRealRizeo:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5039

Removes the duplicate listing title from the excerpt view


garretlaxton:
It looks like the wrong title was removed here. The smaller title is still here, but it's not the one we want.
Screenshot 2022-02-17 at 14-46-24 BD QA – Just another WordPress site

The title should be at the top as seen here.
154341153-7a05e640-14ce-484c-b0bd-59be4a328f77

theRealRizeo:
Thanks @garretlaxton , CSS was off. Should be good now :)

garretlaxton (approved):
Looks great now!

PR #135, opened 2022-02-21
Separate script loading to avoid overloading other non-bd pages

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5041

Currently all BD scripts are loaded on all admin pages in the check to create missing pages. This brings about script conflicts with other plugins and loads BD scripts in places that should not be.
This PR introduces a separate script called `admin-install.js` that is used to only include functions to create new BD pages

Testing :

To test, the notice `Business Directory Plugin requires a page with the [businessdirectory] shortcode to function properly.` should work on all pages in the admin side. On non-bd pages, only the script `wpbdp-admin-install-js` should be registered and visible on Query monitor from the plugin (apart from any other addons which still need to be optimized on script loading)

Notice:

image

Scripts

image

image


includes/admin/class-admin.php
```diff
@@ -1294,7 +1294,7 @@ public function enqueue_scripts( $force = false ) {
 				return;
 			}
 
-			$wpbdp->assets->enqueue_admin_scripts( $force );
+			$wpbdp->assets->register_installation_resources();
```
stephywells, line 1297: It looks like this should load the enqueue_admin_scripts. Maybe we can just remove the force?

includes/class-assets.php
```diff
@@ -267,8 +267,8 @@ public function enqueue_css_override() {
      *
      * @param bool $force Force reloading the resources.
      */
-	public function enqueue_admin_scripts( $force = false ) {
-		if ( ! $force && ! WPBDP_App_Helper::is_bd_page() ) {
```
stephywells, line 271: We still may need to keep the force option for reverse compatibility.

includes/class-assets.php
```diff
@@ -290,6 +290,8 @@ public function enqueue_admin_scripts( $force = false ) {
 
         wp_enqueue_style( 'wpbdp-js-select2-css' );
 
+		do_action( 'wpbdp_enqueue_admin_scripts' );
```
stephywells, line 293: Let’s add a doc block here that includes since. This will make it easier to know if we can use it in addons.

stephywells (changes requested):
Nice! I haven’t tested this yet, but I’ve got a couple requests. It looks good!

garretlaxton (approved):
Looks good, all scripts in the core plugin are removed from all non BD pages in the backend.

includes/class-assets.php
theRealRizeo: The `$force` would bring issues in later on. The [admin_enqueue_scripts](https://developer.wordpress.org/reference/hooks/admin_enqueue_scripts/) uses the string param `$hook_suffix` . We need to only load resources for BD pages

stephywells: There might be pages for an add-on that aren't counted as a Bd page. If there are cases like that, this would break the add-on (or a custom add-on)

stephywells: This should still work fine for an empty string:
if ( ! $force && ! WPBDP_App_Helper::is_bd_page() ) {

stephywells: I updated the message so we can merge this.

PR #136, opened 2022-02-22
Listing image alignment

theRealRizeo:
Reported issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5043

This fixes the listing image alignment and removes the 0 margin on the left and right

image


garretlaxton (approved):
Looks good!

assets/css/less/wpbdp.less
```diff
@@ -152,7 +152,7 @@ form#wpbdmsearchform a.advanced-search-link {
 
 .wpbdp-listing-excerpt .listing-thumbnail {
     float: right;
-    margin: 0 0 0 10px;
+    margin: 0 auto;
```
stephywells, line 155: It looks like this centers the image one smaller screens when it should be aligned left. I think margin:0 looks like it would work better.

stephywells (changes requested):

PR #137, opened 2022-02-22
Single view listing actions spacing

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5044

This adds a space between buttons and actions in the single listing view.

image


stephywells (commented):
Can you tell me how this is different from this?
https://github.com/Strategy11/business-directory-plugin/blob/d5c549efc617717803f813f432e6ef2878dd22af/assets/css/less/wpbdp.less#L115

Do you have a before screenshot too?

garretlaxton (approved):
Looks good!

Before:
155083106-5ca3a5ed-c5bb-4bce-aa84-5bd4983a8220

After: 
Screenshot 2022-02-22 at 14-48-03 Business Real Estate Professionals Directory - BD QA

stephywells:
Thanks for the screenshots Garret.

@theRealRizeo I think we need to do a bit more here. It looks like part of the problem is that there's a ton of duplication here:
https://github.com/Strategy11/business-directory-plugin/blob/master/templates/parts/listing-buttons.tpl.php

Let's get rid of the duplication. Instead of adding a css workaround, it looks like fixing this file will solve the problem. The links will be the same and correctly spaced if they all use the .button class.

theRealRizeo:
Thanks @stephywells I forgot about that. I have added the classes and it looks like this:

image

This is the before view

image

The CSS has been applied to exclude child classes that have the `.wpbdp-button` class as there are still addons that hook buttons which could affect the spacing of the buttons

stephywells:
@theRealRizeo It looks like all the duplicate code is still included in the listing-buttons template. Can you please take a look at this?

theRealRizeo:
Hey @stephywells I have simplified the code. As for places where buttons can be injected, there are some addons that add their buttons as here https://github.com/search?q=org%3AStrategy11+wpbdp-listing-buttons&type=code mainly Claim listings addon

theRealRizeo:
Removed the classes and added a PR for claim listings https://github.com/Strategy11/business-directory-claim-listings/pull/13

stephywells:
Thanks, that's much better. It looks like there could still be an issue if links as added without being a button. Here's the css for the buttons:

```
.listing-actions a.button {
    padding: 5px 10px;
    font-size: 11px;
    text-decoration: none;
    margin-right: 3px;
}
```

Maybe something like this would help for consistency? I haven't looked at this in the browser, just throwing it out there as an example.
```
 .listing-actions > span {
    font-size: 11px;
    margin-right: 10px;
}
```

stephywells:
@theRealRizeo the problem with that is it's a bit too specific, so it's more difficult for themes to override. Also, .back-to-dir-buttons doesn't have any other css.

theRealRizeo:
@stephywells I removed the float on the `.back-to-dir-buttons` as it made no sense and applied the css:

```
.listing-actions > span {
    font-size: 11px;
    margin-right: 10px;
}
```

The output looks neater with standardized font sizes:

image

stephywells (approved):
Thanks, this looks good now.

PR #138, opened 2022-02-23
Include default plan

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5042

Current situation:

Default plan when enabled will not show if payments are possible i.e. if payment gateways are enabled (this is how it worked before the payments switch removal). Other plans would show except the default fee plan

New change:
Default plan will show when enabled regardless of enabled or disabled status. I maintained the args hooks incase a custom addon is using them as we do not want to change the functionality


garretlaxton:
The default fee plan functions great now. However, if no payment gateways are enabled the paid plans remain active, which isn't what our notification says.
Screenshot 2022-02-24 at 15-13-19 Fee Plans ‹ BD QA — WordPress
I think we should add this notification to the **Payments** page as well as it's only visible on the **Fee Plans** page. That way when users do disable their gateways, they will know that the paid plans will also be disabled.

theRealRizeo:
@garretlaxton message has been included in the Payments tab

garretlaxton:
The notification on the payment page looks great. Thanks!

However, I was also trying to point out that paid fee plans still remain active when no payment gateways are enabled. As the notification says `Until you do this, only free plans will be available.`  which isn't true. As you can see the paid plans are still active.
Screenshot 2022-02-25 at 16-49-34 Add Listing »

Paid plans should be automatically get disabled if there are no gateways enabled.

theRealRizeo:
Thanks for the clarification @garretlaxton , I just pushed a change for this, please check

garretlaxton (approved):
Works great now!

includes/helpers/functions/general.php
```diff
@@ -948,12 +948,15 @@ function wpbdp_get_fee_plans( $args = array() ) {
     }
 
 	if ( ! $args['admin_view'] && $args['include_free'] ) {
-		$where .= $wpdb->prepare( ' AND p.amount = %d', 0 );
+		if ( $payments_on ) {
+			$where .= $wpdb->prepare( ' AND p.amount >= %d', 0 );
```
stephywells, line 952: It looks like adding this makes the query more complicated and slower. Also, it changes the behavior. Would this work instead?
```
if ( ! $args['admin_view'] && $args['include_free'] && ! $payments_on ) {
$where .= $wpdb->prepare( ' AND p.amount = %d', 0 );
} ...
```

stephywells (changes requested):
@theRealRizeo can you also update the unit tests to catch this? It seems both the original issue, and the issue Garret caught should be things the unit tests are catching. It's critical that the correct plans are showing up.

theRealRizeo:
@stephywells added some code changes and test cases for the fee plans.

- For the `wpbdp_get_fee_plans` function, the cache key would remain the same regardless of payment status. I noticed this running tests that the results with payment enabled or disabled was always the same (even though there are paid plans) . I added an extra argument after the parsing for `payments_on` that is used to get the plans based on payment status also for a unique cache key
- I added test cases that check for the plans returned when a payment gateway is enabled, a payment gateway is disabled and the default plan is disabled but a payment gateway is enabled

stephywells (commented):
I just made some updates. That part in the code seems to just be adding complication when we are using the status now instead of the amount or tag.

But it looks like the unit test could use a bit of simplification, and use more code from the plugin instead of adding duplication that won't stay up to date.

tests/wpunit/Fee/PlanListingTest.php
```diff
@@ -0,0 +1,144 @@
...
+		$this->tester->wantToTest( 'Test Fee Plan Visibility' );
+		$this->createFees();
+		wpbdp_set_option( 'payments-test-mode', true );
+		wpbdp_set_option( 'authorize-net', 0 );
+		$this->assertFalse( 1 === wpbdp_get_option( 'authorize-net' ), 'Gateway Disabled' );
+		// Random details for testing purposes. We won't attempt a charge, we just need to enable payments.
+		wpbdp_set_option( 'authorize-net-login-id', '7h6MbLNyn9qb' );
+		wpbdp_set_option( 'authorize-net-transaction-key', '98GHnS594xy32V7d' );
+		wpbdp_get_fee_plans( array( 'admin_view' => true ) );
+		$plans = wpbdp_get_fee_plans( array( 'include_free' => true ) );
```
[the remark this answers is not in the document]
theRealRizeo: `get_available_plans` would be in a test case to test listing submission flow in the controllers. This test mainly tests the setting of payment setting and plans availability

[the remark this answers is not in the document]
theRealRizeo: Ideally each function should have its own test case as well with the predefined settings to ensure correct output

theRealRizeo:
@stephywells I have made some changes to the test cases and added the check for `payments_on` in the `wpbdp_get_fee_plans` as all plans were being returned even if payments were not enabled.

For the unit test cases, I agree but in this case it is a test case of the functionality of what is returned with the function `wpbdp_get_fee_plans` with different parameters and settings. Each function used ideally should have its own separate test case as in TDD where in most cases tests are written for the code before its implemented. When the functions pass the test cases, it is easier to gauge how it will react and act in the implementation somewhere else in the plugin. Limiting the test case to use an implementation does not tell us how it will work somewhere else.

I think using implementations of function within controllers would be better as a separate test case where the entire front-end submission flow is emulated using functions in the `WPBDP__Views__Submit_Listing` class

tests/wpunit/Fee/PlanListingTest.php
```diff
@@ -0,0 +1,181 @@
...
+ * Tests for fee Listing
+ */
+class PlanListingTest extends WPUnitTestCase {
+
+	/**
+	 * @var \WpunitTester
+	 */
+	protected $tester;
+
+	public function testFeePlanVisibility() {
```
stephywells, line 22: Let's stick with underscore function names instead of camelcase

tests/wpunit/Fee/PlanListingTest.php
```diff
@@ -0,0 +1,181 @@
...
+		$plans = wpbdp_get_fee_plans();
+		$total = 0;
+		$free_plan = null;
+		foreach ( $plans as $plan ) {
+			$total += absint( $plan->amount );
+			if ( 'free' === $plan->tag ) {
+				$free_plan = $plan;
+			}
+		}
+		$this->assertTrue( ( $total === 0 ), 'Plan total amount is 0' );
```
stephywells, line 64: Should we add an expected plan count in both of these tests? And maybe check if the 'third_free' tag exists also?

stephywells (changes requested):
That's a great improvement! And thanks for much for catching my mistake!!!

tests/wpunit/Fee/PlanListingTest.php
theRealRizeo: Agreed

tests/wpunit/Fee/PlanListingTest.php
theRealRizeo: @stephywells on the unit test cases, it must be camel case

stephywells (approved):
Awesome! This looks great now. Thanks so much for adding those unit tests. That will be a huge help to prevent this issue from coming up any more.

PR #139, opened 2022-02-28
Fix listing image sort order in frontend

theRealRizeo:
Reported issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5045

Fix the image sorting in the frontend. When images on edit listing were drag dropped, the order was not maintained


garretlaxton (approved):
Image ordering works great now!
Screenshot 2022-03-01 at 03-56-46 image order – test site
Screenshot 2022-03-01 at 03-57-12 image order – test site

includes/controllers/pages/class-submit-listing.php
```diff
@@ -1151,8 +1151,8 @@ private function listing_images() {
 				$updated_meta = wpbdp_get_var( array( 'param' => 'images_meta' ), 'post' );
 				$updated_meta = ! empty( $updated_meta[ $img_id ] ) ? (array) $updated_meta[ $img_id ] : array();
 
-				$new_order = ! empty( $updated_meta['order'] ) ? intval( $updated_meta['order'] ) : $order;
-				update_post_meta( $img_id, '_wpbdp_image_weight', $thumbnail_id ? 0 : $new_order );
+				$new_order = ! empty( $updated_meta['order'] ) ? absint( $updated_meta['order'] ) : $order;
+				update_post_meta( $img_id, '_wpbdp_image_weight', $new_order );
```
stephywells, line 1155: Does the thumbnail still get set correctly? If it's working, it looks like we don't need the $thumbnail_id lines above (1133-1139).

stephywells (changes requested):

includes/controllers/pages/class-submit-listing.php
theRealRizeo: Line `1133-1137` have been maintained to set the thumbnail on post. The other `else` has been removed. The thubmnail is set correctly in the first `if` statement

stephywells: Thanks!

stephywells (approved):
Awesome! This looks great!

PR #141, opened 2022-03-01
Block Theme compatibility

theRealRizeo:
Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5040

Block themes render content differently and hooks fire in a different order. This PR fixes the single listing display. The block theme display for a single listing depends on the theme template, different block themes will render the content differently. 
For the 2022 theme, the featured image and title are displayed at the top of the single listing. This is defined in the template file `twentytwentytwo/templates/page.html` I have attached a replacement that removes the top section in the zip file.
The appearance of the listings on block themes is dependent on the theme templates itself. I have not yet seen a way to override this with a custom block template in the active theme that will maintain the structure

image
[2022-templates-page.zip](https://github.com/Strategy11/business-directory-plugin/files/8161763/2022-templates-page.zip)


stephywells (approved):
This looks like a great solution. I updated the css a bit for the forms and buttons.

PR #142, opened 2022-03-01
Image upload issue with CSV import

theRealRizeo:
Related issue : https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5047

Current issue : imports with zip images do not get uploaded and hence do not get attached to a listing. This was mainly due to a change in PHP that strictly validates that the file is an upload via a multipart `POST` method

Fix : The fix uses a backup code to handle image uploads, and introduces a new function `WPBDP__Utils::attach_image_to_media_library` to attach the upload to the media library. This new function returns an attachment_id that is used in the listing.

Testing : There is a sample CSV and zip file in the reported issue that I used for testing. The image should appear in the media library as well as be attached to the listing


garretlaxton (approved):
Looks good! Images are now uploaded on imports.

stephywells (changes requested):
Nice job on this. My suggestions are all really minor.

stephywells (approved):
Thanks! I never got a review request after changes were made.

PR #152, opened 2022-03-07
Better check for categories in loop query

theRealRizeo:
Fix issue with `[businessdirectory-listings]`. Fixed the check for categories in loop to avoid unwanted query args.

Related issue https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5052


theRealRizeo:
@garretlaxton thanks, I have fixed this :) Please check

garretlaxton:
Looking better! I am still having the issue when using a parameter. It should only show Agents but other categories are shown still.
Screenshot 2022-03-08 at 13-23-36 p1 - BD QA

theRealRizeo:
@garretlaxton sorry about that, pushed a fix for it. Please check

garretlaxton (approved):
Works great now! (:

includes/class-shortcodes.php
```diff
@@ -444,17 +444,17 @@ public function sc_listings_featured( $atts ) {
 	 * @since x.x
 	 */
 	private function process_category_atts( $atts, &$query_args ) {
-		if ( ! isset( $atts['category'] ) && ! isset( $atts['categories'] ) ) {
+		if ( ! $atts['category'] && ! $atts['categories'] ) {
```
stephywells, line 447: It looks like 'categories' won't always be set so we'll still need to check for it to avoid php warnings.

includes/controllers/pages/class-all-listings.php
```diff
@@ -39,8 +39,8 @@ public function dispatch() {
             $q->max_num_pages = 1;
         }
         wpbdp_push_query( $q );
-
-        $show_menu = isset( $this->menu ) ? $this->menu : ( ! empty ( $args['tax_query'] ) ? false : true );
+		$show_all = isset( $this->in_shortcode ) ? ! $this->in_shortcode : ( ! empty ( $args['tax_query'] ) ? false : true );
```
stephywells, line 42: This is not very readable. Can this be cleaned up a bit to make it easier to follow? Maybe we can reverse this so we can get rid of the ! and " ? false : true"

stephywells (changes requested):

includes/class-shortcodes.php
theRealRizeo: I have added the array checks and placed some conditional checks to prevent returning or setting values if nothing is returned

includes/class-shortcodes.php
```diff
@@ -450,14 +450,18 @@ private function process_category_atts( $atts, &$query_args ) {
 
 		$requested_categories = array();
 
-		if ( isset( $atts['category'] ) ) {
+		if ( $atts['category'] ) {
 			$requested_categories = array_merge( $requested_categories, explode( ',', $atts['category'] ) );
 		}
 
-		if ( isset( $atts['categories'] ) ) {
+		if ( $atts['categories'] ) {
```
stephywells, line 457: To avoid php warnings, this will need to be:
if ( isset( $atts['categories'] ) && $atts['categories'] ) {

theRealRizeo:
@garretlaxton I did some code changes for the tags as well. Please check and confirm functionality

includes/class-shortcodes.php
```diff
@@ -492,6 +481,43 @@ private function process_category_atts( $atts, &$query_args ) {
...
+	 * @since x.x
+	 */
+	private function process_tag_atts( $atts, &$query_args ) {
+		if ( ! isset( $atts['tag'] ) && ! isset( $atts['tags'] ) ) {
+			return;
+		}
+
+		$requested_tags = array();
+
+		if ( isset( $atts['tag'] ) && $atts['tag'] ) {
```
stephywells, line 500: Now that you point it out... It looks like this is duplicate code. This combines tag and tags, and the other combines category and categories. Maybe we can simplify this?

theRealRizeo: Changes made, combined the check into one reusable function

stephywells: Thanks! It looks like we could make this even less repetitive if we merge them before exploding. Like:
```
$tag = isset( $atts[ $type ] ) ? $atts[ $type ] : '';
$tags = isset( $atts[ $types ] ) ? $atts[ $types ] : '';
$sep = $tag && $tags ? ',' : '';
$tags = $tag . $sep . $tags;
```

theRealRizeo: Thanks :) I have implemented further simplifications on the function

stephywells: Thanks. That doesn't look quite right though. The idea here is to combine 'category' and 'categories'. But the way this was changed would add a trailing or leading comma.

theRealRizeo: Pushed an updated to remove trailing and leading commas. Run a few tests and it the results returned are correct. Not sure how to further simplify it without affecting the expected results

stephywells: This looks like a lot of extra code for this. Was my example not working?

theRealRizeo: The example you shared would not show any listings if no category or tag is passed as a parameter. There would need to be more conditional `if` `else` statements to check if one of the parameters is set. I will make more changes to simplify it

stephywells: Maybe I'm missing something. My example sets both then combines them. If both have a value, it adds a ,

garretlaxton:
Is this ready for QA? I see some changes are still being made.

theRealRizeo:
@garretlaxton changed the label, working on a few things :)

theRealRizeo:
@garretlaxton please test :) ready for QA

includes/class-shortcodes.php
theRealRizeo: Added a `return` if both are not set. The code works now

garretlaxton (approved):
Looks good! Both tag and category parameters.

stephywells (approved):
Looks good now. I simplified this a bit more. Thanks!

PR #153, opened 2022-03-08
Disable listing date edit when it has a subscription

theRealRizeo:
Related issue : https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5053

This disabled the ability to edit the "Expire date" in a listing with an active subscription. Changing the expiration date can bring about many un-supported unrelated issues which we do not support now that include:

- An admin will assume editing the date to a future date for a listing of any status will give/take access to the end user and later on continue with the subscription process at this point
- Editing the date will be assumed to change the gateways that support subscription updates dates

image

Are there any other checks we should add? Subscription status are updated by the gateway

### DIFF
--- a/includes/models/class-listing-subscription.php
+++ b/includes/models/class-listing-subscription.php
@@ -158,7 +158,7 @@ class WPBDP__Listing_Subscription {
             ),
             array( 'listing_id' => $this->listing_id )
         );
-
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
Looks good. Can no longer change the end date of recurring listings.

stephywells (approved):
This looks good. Thanks for using the caching layer!

PR #156, opened 2022-03-14
Re-usable functions as utils

theRealRizeo:
https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5046

Adds reusable functions are utils


stephywells (changes requested):
Is there any place in the core plugin that can use this function? If not, I don't think there's a reason to add it.

PR #157, opened 2022-03-15
Code simplifications

theRealRizeo:
First step on code simplifications and performance enhancements. Would be using the same branch for the changes and submitting new PRs based on it

This first change handles code simplifications to the `payment.php` class to leverage code reuse and the utils cache for queries

### DIFF
--- a/includes/payment.php
+++ b/includes/payment.php
@@ -261,6 +261,9 @@ class WPBDP_PaymentsAPI {
                 continue;
 
             $payment = WPBDP_Payment::get( $data->id );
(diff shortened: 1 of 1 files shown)


stephywells (changes requested):
It doesn't look like this code is worth spending time on. All of these changes are in deprecated code. The supported code is here:
https://github.com/Strategy11/business-directory-premium

It looks like I missed the abandonment_admin_views and abandonment_admin_filter functions though. Would you mind adding those in the Premium plugin?

PR #158, opened 2022-03-15
Remove loading of JQuery UI CSS from CDN

theRealRizeo:
JQuery UI CSS was loading from an external CDN. This PR aims to include only required styling for elements used and deprecates the function `wpbdp_enqueue_jquery_ui_style()`.
Pages affected will be :

- Editing the `Expire Date` in a listing admin view
- Editing a listing in the frontend that has a `Date` Field
- CSV Import with `Select a default user to be used if the username column is not present in the CSV file.` checked for the user list autocomplete
- Modal delete prompt in `Form Fields` page

New styling:

Listing Expire Date admin:
image

Listing Date Field frontend:
image

CSV Import:
image

Modal Delete Prompt:
image


stephywells (changes requested):
It looks like this could still use some work. It's not looking great. We can just copy the css from Formidable (it's only 49 lines). The colors for the front-end should be based on the button color setting in BD.

Here's how it looks with Formidable installed:
Screen Shot 2022-03-15 at 3 12 29 PM

Here's the way it should look (we don't have to switch to dropdowns though):
date_ui-calendar

Also, I think it would be better if we address the to do in there, mentioning that jquery-datepicker-ui should only be loaded when it's needed. We could use the wpbdp_enqueue_jquery_ui_style() function for this, and use wp_enqueue_script( 'jquery-ui-datepicker' ); in that function.

Also, can you help me understand what the new javascript is for in this PR?

theRealRizeo:
The new JS mainly is for the `Today` on click to populate the input field with the current date

theRealRizeo:
Formidable seems to load a custom jQuery UI lib https://github.com/Strategy11/formidable-pro/tree/master/css/ui-lightness
I have added the lib and made some CSS changes

Front Field:
image

Admin backend:
image

theRealRizeo:
@stephywells Thanks for the feedback :) I removed the Jquery UI lib and used the CSS from Formidable. There are CSS elements not fully styled with the Formidable CSS which I added in the `jquery-ui-styles.less` file

garretlaxton:
Looks good, but since we are using the Formidable code here it looks like the 2 are causing some minor front-end issues when both are installed.

When I activate Formidable, the dates on the front-end have some border issues on the bottom and right side.
Screenshot 2022-03-17 at 08-54-09 Add Listing – 0xB9 Sec

And on the receipt page, the CSS for the `<tr>` & `<td>` tags are included in the table here.
Screenshot 2022-03-17 at 08-55-22 Add Listing – 0xB9 Sec

theRealRizeo:
Thanks @garretlaxton I'll add some conditional checks to see if the CSS is loaded :)

theRealRizeo:
@garretlaxton pushed a change to handle this. I do not have a Formidable license so I can't check the interaction well

garretlaxton:
I'm seeing `margin: 0 0 0.4em;` in the `.ui-datepicker table` class of this file here [/assets/css/jquery-ui-styles.min.css](https://github.com/Strategy11/business-directory-plugin/blob/f9e7ead07e87998d092adc0fae70b74b43b70f70/assets/css/jquery-ui-styles.min.css) causing the bottom margin issue.

theRealRizeo:
Calendar in backend has been adjusted to look better

image

Padding issue fixed in frontend, thanks @garretlaxton 

image

garretlaxton (approved):
Looks great!

stephywells:
@garretlaxton would you mind taking one more look over this? I've made some updates so the datepicker script is only loaded on the front-end when there's a date field in the form.

garretlaxton:
@stephywells Looks good. I only see `ui-datepicker` loaded in when a date field is in a form on the frontend.

PR #160, opened 2022-03-16
Remove abandonment hooks

theRealRizeo:
This removes the abandonment hooks migrated to https://github.com/Strategy11/business-directory-premium/pull/25

### DIFF
--- a/includes/payment.php
+++ b/includes/payment.php
@@ -17,10 +17,6 @@ if ( ! class_exists( 'WPBDP_PaymentsAPI' ) ) {
 class WPBDP_PaymentsAPI {
 
     public function __construct() {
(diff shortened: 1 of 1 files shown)


stephywells (changes requested):
Thanks!

We'll need to leave all of the public functions to prevent fatal errors with possible reverse compatibility. But they can all be empty with just the deprecation message

theRealRizeo:
@stephywells thanks for the feedback :) I have added the functions and deprecated them

stephywells (approved):
Thanks! That looks great now.

stephywells (approved):
Thanks! That's perfect now.

PR #164, opened 2022-03-17
Notification center

theRealRizeo:
Adds a notification center on the bottom right of the screen. The notifications on this area are meant to be read by the admin or actioned by the admin.
image

Other notifications such as "Form Field updated" are shown as toast notifications at the bottom center of the screen and disappear after 2.5 seconds.
toast-notifications

Upgrade notifications for the Lite version remain at the top. The notification to install required BD pages has been omitted from the notification center as it would work differently on non-bd pages and bd pages. We can look for a better way to do it in the new interface, perhaps a modal on BD pages?

image

image


stephywells (changes requested):
Thanks for the updates! I'm still not seeing messages for a few things:
- When I update a listing from the back end
- Delete a listing claim (this should be done in a way that doesn't require add-ons tp be updated. I don't think we need to manually mark it in every spot. The default needs to be smarter. If it's dismissible, it should be in the notification center. If it's not dismissible, it can be snackbar. These defaults can be overridden, but that would be the exception rather than the rule)

stephywells:
Oh, and I forgot to mention... We can start on the walkthrough in a new branch. Maybe you can look through these designs and write up the pitch?
https://www.figma.com/file/hey3bHjXZhJpGGt277mHbg/Business_directory?node-id=666%3A17092

The pitch should break the tasks down into 1-2 week deliverables. Here's the pitch template to duplicate:
https://drive.google.com/drive/folders/1SOFAPBmAEAnHYFgbr-ja-qj8WIr4Wc8s?ths=true

The technical section should be very detailed, and spell out exactly how you plan to do it, and what you plan on doing. Getting the pitch just right should save a lot of time during development.

theRealRizeo:
@stephywells I have made corrections to the listing edit page.
Regarding the claim listings, it seems this is not working and will have to be looked into in the addon level. I tried with the master branch of core and did not get any notification from claim listings

theRealRizeo:
Notifications have been updated and I have added extra classes to notifications that should be in the notification area.
The JS has been simplified to check `dismissible` notices and add them in the correct place

garretlaxton:
Looks good, but a few questions.

Should the notifications appear on the settings page? 

And if I dismiss a notification and refresh the page the notifications appear again. Is that expected? I feel that could get annoying for users. A timer might be a good idea after a notification is dismissed before showing again.

stephywells:
@garretlaxton the notifications should appear on the same pages as they did before. We need to make sure the notifications show up when we expect them too.

The dismissible notices should be gone once they are dismissed. If you are seeing some that come right back, maybe you can mention specific examples?

theRealRizeo:
I agree, some dismissed notifications should return as some are setting notifications on things that need to be configured

theRealRizeo:
I agree, some dismissed notifications should return as some are setting notifications on things that need to be configured

theRealRizeo:
I agree, some dismissed notifications should return as some are setting notifications on things that need to be configured

garretlaxton:
Gotcha, one separate notification I can't dismiss is when I delete a listing. You can see this by moving a listing to trash or permanently deleting it.
Screenshot 2022-03-18 at 06-52-26 Directory ‹ test site — WordPress

stephywells:
Looks like Paul added fix for that. Are you finding any other issues @garretlaxton ?

garretlaxton:
Don't see anything else related to the changed here. 

One separate issue I found is when you create a listing in the back-end, leave everything blank and click Publish. A notification shows saying Post published which isn't true. This false notification now overlaps the real notification in the new UI.
Screenshot 2022-03-18 at 20-44-12 Edit Listing “Auto Draft” ‹ test site — WordPress

theRealRizeo:
The `Post Published` is from WordPress when it saves as Draft. I will check the overlapping issue

garretlaxton:
@theRealRizeo Since that is coming from WordPress, is there a way to disable the publish button until something has been input like WordPress does with posts?
Screenshot 2022-03-21 at 14-56-33 Add New Post ‹ BD QA — WordPress

Because as of now it says Post Published, but it actually isn't, and nor is it saved as a draft. I can create a separate ticket for this as it's not really related here, but just wanted to mention it since it was interfering with other notifications.

theRealRizeo:
@garretlaxton It should be possible with some custom JS t disable until the Title and some listing field data are added

stephywells:
@garretlaxton yeah, that is a little weird. Would you mind creating a new issue for this? We'd handle that separately since this part is just about the message placement and style rather than the text.

Does this look better to you now when there are multiple notifications? The problem wasn't about where the message was coming from, but we just didn't account for there being more than one in that area.

garretlaxton:
Looks better, notifications now stack. But the width seems to always be 100%.
Screenshot 2022-03-22 at 04-56-43 Edit Listing “Auto Draft” ‹ 0xB9 Sec — WordPress

Removing
```
.wpbdp-snackbar {
    width: 100%;
}
```
and
```
#wpbdp-snackbar-notices {
    text-align: center;
}
```
Seems to make it fit better. It could probably use some padding for the dismiss notification. 
Screenshot 2022-03-22 at 05-08-57 Edit Listing “Auto Draft” ‹ 0xB9 Sec — WordPress

Not sure if there's a better way to go about it, but that's what I found!

stephywells:
@garretlaxton would you mind clearing your cache and trying that once more? It looks like the css cleared but not the js.

garretlaxton:
@stephywells Unfortunately, I still see the same after clearing cache and viewing in incognito.

PR #182, opened 2022-04-27
Use the h1 from the theme

stephywells:
This issues seems to come up a lot. Now instead of janky hacks to clear out the post title, we'll let the theme show the post title.

This will need more testing and comparison in different WP and BD themes.

Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5062

@garretlaxton I could use some help testing this one. We need to keep this from looking really bad on any sites.


garretlaxton (approved):
From what I see, h1 tags will use the current WP theme style. I tested this with about 10 different themes from the popular tab, also a few common themes I see people using with BD.

Our BD themes only change h2 tags so they look fine as well.

PR #196, opened 2022-05-31
Fixes authentication error

stephywells:
Before: Adding &listing_id=27 to the new listing creation page when logged out can allow editing.

After:
This adds more checks before pulling a listing when not in edit mode.

This also fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/4923

@garretlaxton 
This could use more testing to:
- make sure listings can be submitted when there are multiple pages, when logged in as admin, logged in as subscriber, or logged out
- make sure nothing is reset when going back and forth between pages in the new listing form. This is loading with ajax now, so it's important that it fetches the correct entry each time.
- Test this on the sandbox site with page caching when logged out https://sandbox.formidableforms.com/salman2/wp-admin/edit.php?post_type=wpbdp_listing (I'll get this beta version uploaded now)


stephywells:
@garretlaxton Will you have some time to test this today so we can release it tomorrow?

garretlaxton:
Sorry, I didn't see your comment about this till today. This fix looks good, I don't have access to Salman's sandbox to test there I will need permission. However, something similar I see is our `wpbdp-listing-submit-image-upload` action doesn't have a permission check either.

**Example:** You can upload an image to your listing and it creates a POST request `http://localhost/wp-admin/admin-ajax.php?action=wpbdp-listing-submit-image-upload&listing_id=249&_wpnonce=1f2b99d766`. If you change the `&listing_id=249` to another existing listing ID you can add an image to that listing. The POST request also includes a `images_count` parameter which you can set to `0` and bypass the image upload limit set in the fee plan therefore uploading as many images as you want. Not sure if that's something you want to include here or in a separate PR, but I can log it if needed.

stephywells:
@garretlaxton Thanks for testing. You have access now to that site.

Would you mind adding that as a new issue, and remove the details here? I don't really want this public since it isn't fixed yet.

garretlaxton (approved):
Oops, forgot this was a public repo... The QA site looks good when logged out, I also tried including a listing ID.

stephywells:
@garretlaxton I'm not seeing the other issue you spotted logged. Did you create that issue?

PR #200, opened 2022-06-07
Fix category title for most themes

stephywells:
This fix doesn't work for block themes, but solves the problem for others

Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5084


garretlaxton:
@stephywells Is there any solution for this with block themes since you mentioned this doesn't fix those? Someone with Astra theme is seeing this.

PR #205, opened 2022-06-21
Follow page redirects on the submit listing form

stephywells:
This fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5088

This affects a couple spots that should all be tested:
- redirecting to the login page when login is required before submitting a listing
- clicking the "clear form" link in the submit listing form
- redirecting to checkout when there is a pending payment for the listing


garretlaxton:
Required login option looks good now. I do notice when going to the submit listing page as a logged-out user there is a slow load time before the page loads. It goes to the blank page, then loads the correct page. This happens regardless of the required login option being enabled or disabled.

The cleared form button works as normal still, but this restarts the whole process from the first page (category selection). Would it be better if this just cleared the form and didn't restart? Or maybe change the text to `Restart Listing` since that's what it does. Just an idea.

Pending payments redirected great to the selected gateway. I also tested with renewals.

PR #210, opened 2022-07-05
Fix the fatal error on the manual upgrade page

stephywells:
This fixes the error mentioned here:
https://wordpress.org/support/topic/critical-error-after-updating-3/

This can be seen by installing a 4.x version and then update to current.


garretlaxton (approved):
This looks great! I don't see the error anymore when doing the manual upgrade. 

Please check my commit also, the button after the upgrade completed was pointing to `admin.php?page=wpbdp_admin` which we don't use anymore. I changed it to our main directory page.

PR #221, opened 2022-07-19
Fix category header for block themes

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5094
This prevents the title and image showing at the top of the category page in some themes like Astra and 2022


garretlaxton (approved):
This one looks good! I don't see any issues with this.

PR #227, opened 2022-07-26
Process shortcodes in textarea when enabled

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5119

Shortcodes were being processed all the time before 6.2.6, even when the setting was turned off. This correctly processes shortcodes only when they should be processed.


garretlaxton (approved):
This looks good! When the option is enabled the shortcode shows the content, when disabled only the shortcode text shows.
Screenshot 2022-07-26 at 17-01-13 Andrew Agent – BD Demo

PR #229, opened 2022-07-27
More flexible currency options

stephywells:
- Get the currency options in less settings
- If currency isn't included, add a "Custom" option to set a new one
- also fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5122

**Before:**
Screen Shot 2022-07-27 at 4 54 59 PM

**After:**
Conditional custom option:
Screen Shot 2022-07-27 at 4 54 20 PM

More simple option that sets currencies better:
Screen Shot 2022-07-27 at 4 54 44 PM


garretlaxton:
Few questions and issues..

1. Should both Currency Code fields be visible when a currency is selected? It seems like the 2nd field should only show if `Custom` is selected.
Screenshot 2022-07-28 at 00-56-09 Settings ‹ BD Demo — WordPress

2. Kinda like 1, when a currency is selected the symbol and symbol display don't seem do anything. Maybe hide these options unless a custom currency is selected?
Screenshot 2022-07-28 at 01-03-31 Settings ‹ BD Demo — WordPress
Screenshot 2022-07-28 at 01-05-52 Plans ‹ BD Demo — WordPress


3. My fees aren't showing the price when I go to create a listing.
Screenshot 2022-07-28 at 00-59-14 Add Listing – Just another WordPress site
Screenshot 2022-07-28 at 01-01-53 Plans ‹ BD Demo — WordPress

stephywells:
Wow. Thanks for catching that!

This is fixed now. Let me know if you see anything else.

garretlaxton (approved):
This looks great now! I checked everywhere I could think of that we display a price and they all format properly.

I also made a successful payment with a custom currency.
Screenshot 2022-07-28 at 22-39-36 BD QA Just another WordPress site

PR #244, opened 2022-08-18
recaptcha fixes

stephywells:
- simplify the recaptcha settings
- use correct labels for keys
- Fixes conflict with Wordpress login and registration form
- fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5099
- fixes conflict with BWS registration recaptcha
- fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5100 (partly)

Conflict still exists with BWS login. This may need to be fixed from their end.

I tested with v3 recaptcha, but not with v2.

### DIFF
--- a/includes/admin/settings/class-settings-bootstrap.php
+++ b/includes/admin/settings/class-settings-bootstrap.php
@@ -104,66 +104,74 @@ final class WPBDP__Settings__Bootstrap {
 
         // reCAPTCHA.
         wpbdp_register_settings_group(
-            'recaptcha',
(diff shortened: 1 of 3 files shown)


garretlaxton (approved):
I'm liking the new look on the spam page! The v2 captcha works great, even with Wordfence & BWS.

I see you went over this with v3, would it be worth me checking it with v3 as well?

stephywells:
Thanks for checking that! That's perfect.

PR #264, opened 2022-10-21
Added Test Mode column in payments to indicate if test mode is enabled

martinjankov:
https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5166

### DIFF
--- a/includes/installer.php
+++ b/includes/installer.php
@@ -10,7 +10,7 @@ require_once WPBDP_PATH . 'includes/admin/upgrades/class-migration.php';
  */
 class WPBDP_Installer {
 
-    const DB_VERSION = '18.6';
+    const DB_VERSION = '18.7';
 
     private $installed_version = null;
 
@@ -128,7 +128,7 @@ class WPBDP_Installer {
             currency_code varchar(3) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL DEFAULT 'USD',
             amount decimal(10,2) NOT NULL DEFAULT 0.00,
             status varchar(255) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL,
-            is_test boolean NULL DEFAULT FALSE,
+            is_test boolean NULL,
             created_at datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
             processed_on timestamp NULL DEFAULT NULL,
             processed_by varchar(255) CHARACTER SET utf8 COLLATE utf8_general_ci NULL,
--- a/includes/installer.php
+++ b/includes/installer.php
@@ -10,7 +10,7 @@ require_once WPBDP_PATH . 'includes/admin/upgrades/class-migration.php';
  */
 class WPBDP_Installer {
 
-    const DB_VERSION = '18.5';
+    const DB_VERSION = '18.6';
 
     private $installed_version = null;
 
@@ -128,6 +128,7 @@ class WPBDP_Installer {
             currency_code varchar(3) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL DEFAULT 'USD',
             amount decimal(10,2) NOT NULL DEFAULT 0.00,
             status varchar(255) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL,
+            is_test boolean NULL DEFAULT FALSE,
             created_at datetime NOT NULL DEFAULT '0000-00-00 00:00:00',
             processed_on timestamp NULL DEFAULT NULL,
             processed_by varchar(255) CHARACTER SET utf8 COLLATE utf8_general_ci NULL,
--- a/includes/models/class-payment.php
+++ b/includes/models/class-payment.php
@@ -380,6 +380,8 @@ class WPBDP_Payment extends WPBDP__DB__Model {
      * @return void
      */
     public function maybe_set_test_mode() {
+        $this->is_test = false;
+
         $testing_mode = wpbdp_get_option( 'payments-test-mode' );
 
         if ( $testing_mode ) {
--- a/includes/models/class-payment.php
+++ b/includes/models/class-payment.php
@@ -380,13 +380,7 @@ class WPBDP_Payment extends WPBDP__DB__Model {
      * @return void
      */
     public function maybe_set_test_mode() {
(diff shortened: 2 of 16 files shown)


includes/installer.php
```diff
@@ -128,6 +128,7 @@ public function get_database_schema() {
             currency_code varchar(3) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL DEFAULT 'USD',
             amount decimal(10,2) NOT NULL DEFAULT 0.00,
             status varchar(255) CHARACTER SET utf8 COLLATE utf8_general_ci NOT NULL,
+            is_test boolean NULL DEFAULT FALSE,
```
stephywells, line 131: I'm thinking we might need to not have a default here. If we do, we won't have a way of knowing if the payment was not in test mode, or if it was created before the DB change.

includes/models/class-payment.php
```diff
@@ -77,6 +77,8 @@ protected function before_save( $new = false ) {
                 $this->status = 'completed';
             }
         }
+        
+        $this->maybe_set_test_mode();
```
stephywells, line 81: I am seeing an issue here. It looks like this is running every time a payment is saved. So when I go to an existing payment and click "save payment", the payment mode gets changed. 

Is there a spot we could add this so it only gets added when the payment is created?

templates/admin/payments-details.tpl.php
```diff
@@ -84,7 +84,7 @@
                                 <div class="wpbdp-admin-box-row">
                                     <label><?php esc_html_e( 'Gateway', 'business-directory-plugin' ); ?></label>
                                     <?php /* translators: Gateway: (Not yet set) */ ?>
-                                    <?php echo esc_html( $payment->gateway ? $payment->gateway : __( '(Not yet set)', 'business-directory-plugin' ) ); ?>
+                                    <?php echo esc_html( $payment->gateway ? sprintf( '%s %s', $payment->gateway, $payment->is_test ? __( '- Test Mode', 'business-directory-plugin' ) : '' )  : sprintf( '%s %s', __( '(Not yet set)', 'business-directory-plugin' ), $payment->is_test ? __( '- Test Mode', 'business-directory-plugin' ) : '' ) ); ?>
```
stephywells, line 87: There seems to be some duplication in here. Maybe we can simplify this? I think it would be fine to add the test mode part on the next line too rather than tacking it on here.

stephywells (approved):
Thanks so much @martinjankov ! This worked great for me!

I did come across one issue. I've also got a couple smaller suggestions.

stephywells (changes requested):
Oops! I didn't mean to mark approved.

includes/installer.php
martinjankov: That makes sense yes. I will update it.

includes/models/class-payment.php
martinjankov: @stephywells I can probably use the $new param that is passed to the before_save method if that presents a new payment. I will check it what exactly is sent there.

templates/admin/payments-details.tpl.php
martinjankov: Next line would be simpler. I just followed the example you had in the issue. I will update

martinjankov:
@stephywells I updated the requested changes

includes/models/class-payment.php
```diff
@@ -370,6 +374,19 @@ public function set_payment_method( $method ) {
...
 
+    /**
+     * Check if payment gateway is in test mode and set the payment as testing if so
+     *
+     * @return void
+     */
+    public function maybe_set_test_mode() {
+        $testing_mode = wpbdp_get_option( 'payments-test-mode' );
+
+        if ( $testing_mode ) {
```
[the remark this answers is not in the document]
stephywells: Yes! I think that would be great to know for sure.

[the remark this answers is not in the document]
martinjankov: @stephywells This is updated with the latest commit.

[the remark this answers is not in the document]
stephywells: Thank you! I think we could simplify this a bit more with:
`$this->is_test = (bool) wpbdp_get_option( 'payments-test-mode' );`

templates/admin/payments-details.tpl.php
```diff
@@ -85,6 +85,12 @@
                                     <label><?php esc_html_e( 'Gateway', 'business-directory-plugin' ); ?></label>
                                     <?php /* translators: Gateway: (Not yet set) */ ?>
                                     <?php echo esc_html( $payment->gateway ? $payment->gateway : __( '(Not yet set)', 'business-directory-plugin' ) ); ?>
+                                    <?php
+                                        printf(
+                                            '<span class="wpbdp-payment-test-mode-label">%s</span>',
```
stephywells, line 90: This would be more consistent with the existing codde if we use inline php inside the html.
```
<span class="wpbdp-payment-test-mode-label">
<?php stuff here ?>
</span>
```

stephywells (changes requested):
Thanks for the update @martinjankov ! I think we can simplify this a bit more before we merge it.

templates/admin/payments-details.tpl.php
martinjankov: @stephywells yea that would be.. I just hate html within php, but for consistency I will change it

martinjankov:
@stephywells Changes are added again

stephywells:
@garretlaxton would you mind looking this over? Here's a bit more detail about what is changing here:
https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5166

It would be great to make sure the "test mode" label shows up for new test payments, and doesn't show for live payments. This label should also not change when a payment is updated.

garretlaxton:
When I submit a recurring listing through Stripe in test mode, I notice 2 payments are going through. One shows the test mode label and the other doesn't.
image

I'm still looking.

garretlaxton (approved):
This looks good for me!

Live
image
Test mode
image

One suggestion though, would it be worth having an attribute that shows `Test Mode`?
image

stephywells:
That's a great idea @garretlaxton . It would definitely be helpful to show test payments on the page that lists the payments too. Thank you!

PR #277, opened 2022-12-22
Improve the speed of accessing the BD Plugin dashboard

shervElmi:
### Issue link:
Fixes https://github.com/Strategy11/business-directory-premium/issues/110

With this Pull Request, we will improve the speed of accessing the BD Plugin dashboard.


shervElmi:
@stephywells I removed the loading of SVG icon content by the WP File System. Actually, I've tried to fix this problem by loading SVG content with other methods but they didn't work, and made the site slow!

stephywells (changes requested):
This works well! The only problem is this makes the icons black. It looks like adding opacity: .9 makes a big difference. Would you mind adding that, and then we can merge this?

PR #279, opened 2022-12-29
Format PHP files

shervElmi:
With this Pull Request, we will format all PHP files.


stephywells (approved):
I didn't review every file line by line, but I reviewed about half. Everything I looked at was great. Thanks for keeping this pr limited to just a few rule updates.

PR #280, opened 2022-12-30
Fix the display listing title issue

shervElmi:
### Issue link:
https://github.com/Strategy11/business-directory-premium/issues/44

With this Pull Request, we will fix the display listing title issue.

### Note:
It is better to add the following code in the [Troubleshooting problems in Business Directory Plugin](https://businessdirectoryplugin.com/knowledge-base/troubleshooting-problems-in-business-directory-plugin/) and after updating the plugin, users who have this problem should run this code.

`add_filter( 'wpbdp_heading_type', '__return_true' );`

@bobbie-d After the new version please send this code to this user:
https://github.com/Strategy11/business-directory-regions/issues/34


garretlaxton (approved):
This looks great when the header title option is set to hidden for Enfold theme. 
image

Let me know if there are any other themes worth testing!

shervElmi:
Thank you @garretlaxton! No, other themes don't have a specific condition to cause this problem and we provide users with the following PHP code:

`add_filter('wpbdp_heading_type', '__return_true');`

templates/parts/listing-title.tpl.php
```diff
@@ -6,12 +6,22 @@
...
-	<div class="listing-title">
+	<?php
+		$class_name = 'listing-title';
+
+		if ( $title_type === true ) {
+			$title_type = 'h1';
+			$class_name .= ' show-listing-title';
+		}
+	?>
+	<div class="<?php echo esc_attr( 'listing-title' . $class_name ); ?>">
```
stephywells, line 17: Will the class end up looking like this?
"listing-titlelisting-title show-listing-title" or "listing-titlelisting-title"

includes/compatibility/class-cpt-compat-mode.php
```diff
@@ -174,4 +175,18 @@ public function enfold_title( $args, $id ) {
...
 		return $args;
 	}
+
+	/**
+	 * Display the BD-plugin title if Enfold header is disabled.
+	 *
+	 * @since x.x
+	 */
+	public function enfold_maybe_add_single_title() {
+		/** @phpstan-ignore-next-line */
```
stephywells, line 185: Why does this need to be ignored?

A lot of this looks like duplication. Is it possible to merge this duplicate code with the theme_enfold_maybe_add_single_title function?

includes/compatibility/class-themes-compat.php
```diff
@@ -271,6 +300,20 @@ public function theme_enfold_tax_title( $args, $id ) {
 		return $args;
 	}
 
+	/**
+	 * Display the BD-plugin title if Enfold header is disabled.
+	 *
+	 * @since x.x
+	 */
+	public function theme_enfold_maybe_add_single_title() {
+		/** @phpstan-ignore-next-line */
```
stephywells, line 309: Can we remove this ignore?

includes/helpers/class-listing-display-helper.php
```diff
@@ -123,6 +123,15 @@ private static function basic_vars( $listing_id ) {
...
 		$vars['title_type'] = apply_filters( 'wpbdp_heading_type', 'h1' );
 
+		/**
+		 * Filters base variables passed to a template.
+		 *
+		 * @since x.x
+		 *
+		 * @param array $vars Array of variables passed to a template.
+		 */
+		$vars = apply_filters( 'wpbdp_template_basic_vars', $vars );
```
stephywells, line 133: Is this being used somewhere?

includes/compatibility/class-cpt-compat-mode.php
```diff
@@ -139,6 +139,7 @@ public function get_title( $title, $id ) {
 		} elseif ( function_exists( 'avia_title' ) ) {
 			// Change the main title in Enfold theme.
 			add_filter( 'avf_title_args', array( &$this, 'enfold_title' ), 10, 2 );
+			$this->enfold_maybe_add_single_title();
```
stephywells, line 142: Is there a way we can handle this in the WPBDP__Themes_Compat class? Maybe we could use the avf_title_args filter in the line above instead of adding an extra function here?

stephywells (changes requested):
This looks like a good start. I think we can improve it a bit before merging.

templates/parts/listing-title.tpl.php
shervElmi: Yes, exactly.

includes/compatibility/class-cpt-compat-mode.php
shervElmi: Yes, of course. I saw that some code was repeated to make it compatible, and I decided to act accordingly so as not to mess up the order.

templates/parts/listing-title.tpl.php
stephywells: From what I can see, listing-title is duplicated and there is a space missing. Are you seeing this too?

includes/compatibility/class-themes-compat.php
shervElmi: I got a GitHub check error and find some codes act like this!

includes/helpers/class-listing-display-helper.php
shervElmi: No, I added this for the future!

includes/compatibility/class-cpt-compat-mode.php
shervElmi: No, there isn't. because avf_title_args executes after the listing-title.tpl.php file.

templates/parts/listing-title.tpl.php
shervElmi: Oh, my bad!

includes/compatibility/class-cpt-compat-mode.php
stephywells: I'm not seeing what the error was, but I'm guessing it was undefined function? If so, we need to add avia_header_setting() into the stubs.php in order to remove this ignore statement.

includes/compatibility/class-themes-compat.php
stephywells: I'm not seeing what the error was, but I'm guessing it was undefined function? If so, we need to add avia_header_setting() into the stubs.php in order to remove this ignore statement.

includes/compatibility/class-cpt-compat-mode.php
stephywells: How about we add a new hook here instead then and move all of this theme code to the other class? Like this:

```
if ( is_main_query() && ( in_the_loop() || $is_block ) && $id === (int) wpbdp_get_page_id( 'main' ) ) {
			$title = $this->page_title;
		}
$title = apply_filters( 'wpbdp_get_title_compat', $title );
```

Then, inside `theme_enfold_before_dispatch()` we can add:
```
// Change the main title in Enfold theme.
add_filter( 'avf_title_args', array( &$this, 'enfold_title' ), 10, 2 );
add_filter( 'wpbdp_get_title_compat', 'new_function_here_that_returns_title...' );
```

shervElmi: Please see the codes below!

image
image


One of the files `class-themes-compat.php` or `class-themes-compat.php` is executed and we cannot connect the codes of these 2 classes together.

includes/compatibility/class-cpt-compat-mode.php
shervElmi: I removed it in this [commit](https://github.com/Strategy11/business-directory-plugin/pull/280/commits/974b4a485676b15d2b7e7267f7e907573d9614ec).

stephywells: It's still here. It looks like it got removed from one function but not the other.

shervElmi: Sorry, I just removed the other one.

stephywells (approved):
This looks good now! Thanks Sherv!

shervElmi:
Thank you Steph!

PR #281, opened 2023-01-09
Fix row size too large errors with InnoDB

shervElmi:
### Issue link:
https://github.com/Strategy11/business-directory-premium/issues/107

If InnoDB strict mode is enabled, and if a table exceeds its row format's [maximum row size](https://mariadb.com/kb/en/innodb-row-formats-overview/#maximum-row-size), then InnoDB will return an error.

```
ERROR 1118 (42000): Row size too large (> 8126). Changing some columns to 
TEXT or BLOB may help. In current row format, BLOB prefix of 0 bytes is stored inline.
```

With this Pull Request, we will add `ROW_FORMAT=DYNAMIC;` to the database schema to prevent this error.

### DIFF
--- a/includes/installer.php
+++ b/includes/installer.php
@@ -91,7 +91,7 @@ class WPBDP_Installer {
             shortname varchar(255) NOT NULL DEFAULT '',
             tag varchar(255) NOT NULL DEFAULT '',
             KEY field_type (field_type)
-        ) DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci;";
+        ) DEFAULT CHARSET=utf8 COLLATE=utf8_general_ci ROW_FORMAT=DYNAMIC;";
 
 		$schema['plans'] = "CREATE TABLE {$wpdb->prefix}wpbdp_plans (
             id bigint(20) PRIMARY KEY  AUTO_INCREMENT,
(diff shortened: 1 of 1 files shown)


stephywells:
@shervElmi do you have more information on what DB columns were returning such large values? The linked issue doesn't look related. Where can I find more detail on the original problem?

shervElmi:
@stephywells I was trying to resolve this issue on the user's site and for testing, I added the is_test column to the payments table, I got this error: Row size too large...

To find it, we can use bash as mentioned in the [MongoDB documentation](https://mariadb.com/kb/en/troubleshooting-row-size-too-large-errors-with-innodb/#finding-all-tables-that-currently-have-the-problem).

But I have already fixed the user's problems and we can't test this issue on their sites. Let me see if I can create an environment in localhost that will encounter this problem so that I can give you an accurate answer.

stephywells:
@shervElmi This concerns me to add this as a blanket across all tables and all columns if we don't really know what effect if may have or where the problem came from. I would feel much better about this if we could replicate the issue and narrow it down so we know why there was a huge amount if data in one row.

shervElmi:
@stephywells In fact, I committed to talk to you about this, but the quick solution is this, but I am also concerned! and the other solution is to refactor the tables to smaller tables, which we still do not have a valid reason to do.

I know the solutions and my main suggestion is that if a user faces this problem, we implement reliable solutions on their staging site so that we can make the right decision.

stephywells:
That sounds like a good plan to me. If this comes up again, let's get to the bottom of it. I'm seeing three columns in the payments table that allow for lots of data.
```
payment_items longblob NULL,
data longblob NULL,
payer_data blob NULL,
```

I'm not really sure what could be here that would be so big though. Let's plan on researching this more if it comes up again.

PR #283, opened 2023-01-11
Fix showing two admin menus for the editor role

shervElmi:
### Issue link:
https://github.com/Strategy11/business-directory-premium/issues/106

Changes proposed in this Pull Request:
With this Pull Request, we will fix showing two admin menus for the editor role.

#### Before PR: 
image

#### After PR:
image

### DIFF
--- a/includes/admin/class-admin.php
+++ b/includes/admin/class-admin.php
@@ -515,9 +515,16 @@ if ( ! class_exists( 'WPBDP_Admin' ) ) {
 			$cpt_menu   = 'edit.php?post_type=' . WPBDP_POST_TYPE;
 			$admin_menu = $this->menu_id;
 
-			if ( isset( $submenu[ $cpt_menu ] ) && isset( $submenu[ $admin_menu ] ) ) {
+			$current_user = wp_get_current_user();
+			$roles        = $current_user instanceof WP_User ? $current_user->roles : [];
+			$role         = ! empty( $roles ) && is_array( $roles ) ? array_shift( $roles ) : '';
+
+			if ( isset( $submenu[ $cpt_menu ] ) && ( isset( $submenu[ $admin_menu ] ) || $role === 'editor' ) ) {
 				$this->change_menu_name( $submenu[ $cpt_menu ] );
-				$submenu[ $admin_menu ] = array_merge( $submenu[ $cpt_menu ], $submenu[ $admin_menu ] );
+
+				$submenu[ $admin_menu ] = isset( $submenu[ $admin_menu ] ) ?
+					array_merge( $submenu[ $cpt_menu ], $submenu[ $admin_menu ] ) :
+					$submenu[ $cpt_menu ];
 			}
 		}
 


includes/admin/class-admin.php
```diff
@@ -515,9 +515,16 @@ public function admin_menu_combine() {
 			$cpt_menu   = 'edit.php?post_type=' . WPBDP_POST_TYPE;
 			$admin_menu = $this->menu_id;
 
-			if ( isset( $submenu[ $cpt_menu ] ) && isset( $submenu[ $admin_menu ] ) ) {
+			$current_user = wp_get_current_user();
+			$roles        = $current_user->roles;
+			$role         = ! empty( $roles ) && is_array( $roles ) ? array_shift( $roles ) : '';
```
stephywells, line 520: I think this would be a lot cleaner if we can use a capability instead of a role. Here's a list of options. I'm thinking [edit_others_pages](https://wordpress.org/support/article/roles-and-capabilities/#edit_others_pages) would work well here.

includes/admin/class-admin.php
```diff
@@ -515,9 +515,16 @@ public function admin_menu_combine() {
...
+
+			if ( isset( $submenu[ $cpt_menu ] ) && ( isset( $submenu[ $admin_menu ] ) || $role === 'editor' ) ) {
```
stephywells, line 522: This looks like it could potentially have unwanted side effects. Can we move this to elseif instead?

This issue has been repeating. Can you also add unit tests to check for the expected menu items for different user roles?

stephywells (changes requested):
This looks good, but I'm not very confident that this won't cause issues. I think we can make this a bit more clear to make sure it only changes menus when we want them to change.

includes/admin/class-admin.php
shervElmi: @stephywells I revert the previous codes and just committed the new codes.

shervElmi:
@stephywells I'll create another PR for updating TESTS.md and creating a `.env.tests `that make it easy to run tests on local machines, but now I have other high-priority tasks.

garretlaxton:
Could we add some padding for the editor role? The items on the sides are on the edge of the screen, and the search bar is almost hidden by the Screen Options setting.
image

Also, should editors be able to add categories and tags like they can with normal posts? There's no option for that currently.

shervElmi:
> Could we add some padding for the editor role? The items on the sides are on the edge of the screen, and the search bar is almost hidden by the Screen Options setting. image
> 
> Also, should editors be able to add categories and tags like they can with normal posts? There's no option for that currently.

@garretlaxton It's another issue and not specifically for the editor user.
Would you mind creating another issue for it?

stephywells (changes requested):
@shervElmi I'm not seeing anything in that test that would cover this specific issue. There should be a test for editor role users that would fail without this change, and pass with it. Here's what I would really like the unit tests to cover:
1. As admin, check to make sure the 'wpbdp_admin' menu exists and wp-admin/edit.php?post_type=wpbdp_listing does not exist at the top level.
2. As editor, run the same test 
3. As admin, check for wp-admin/edit.php?post_type=wpbdp_listing submenu, wpbdp_settings, wpbdp_admin_payments, wpbdp-addons, and wpbdp-themes.
4. As editor, check to make sure the 404 url mentioned in the issue report does not exist https://github.com/Strategy11/business-directory-premium/issues/106

It looks like this fix went the wrong direction and the wpbdp_admin doesn't exist for editors. I went ahead and added a commit to cover the open issues for editors.

tests/wpunit/Admin/AdminTest.php
```diff
@@ -0,0 +1,135 @@
...
+		wp_set_current_user( $user_id );
+		static::$admin->admin_menu();
+		static::$admin->hide_menu();
+
+		$this->assertArrayHasKey( static::$menu_id, $submenu );
+
+		$submenu = $original_submenu;
+	}
+
+	protected function wpbdp_listing_post_type_at_top_level_exists( $user_id ) {
```
shervElmi, line 129: @stephywells I couldn't find a way to check Listing Post Type as a top-level menu! Do you have any suggestions?

tests/wpunit/Admin/AdminTest.php
```diff
@@ -0,0 +1,135 @@
...
+		$this->tester->wantToTest( 'Admin Menu Exists For Editor User' );
+		$this->wpbdp_admin_menu_exists( static::$editor_user_id );
+	}
+
+	public function testListingPostTypeDoesNotExistAtTopLevelAsAdmin() {
+		$this->tester->wantToTest( 'Listing Post Type Does Not Exist At Top Level As Admin' );
+		$this->wpbdp_listing_post_type_at_top_level_exists( static::$admin_user_id );
+	}
+
+	public function testCombineListingPostTypeAndWPBDP_Admin() {
```
shervElmi, line 84: @stephywells I'll complete this test after finding a way to check Listing Post Type as a top-level menu.

tests/wpunit/Admin/AdminTest.php
stephywells: Is this included in $menu? What does the menu include with var_dump($menu)?

shervElmi: It returns this array:

```
Array
  (
      [20] => Array
          (
              [0] => Directory
              [1] => edit_posts
              [2] => wpbdp_admin
              [3] => Business Directory Admin
              [4] => menu-top toplevel_page_wpbdp_admin
              [5] => toplevel_page_wpbdp_admin
              [6] => ...
          )
  
      [20.06409] => Array
          (
              [0] => Directory
              [1] => edit_posts
              [2] => wpbdp_admin
              [3] => Business Directory Admin
              [4] => menu-top toplevel_page_wpbdp_admin
              [5] => toplevel_page_wpbdp_admin
              [6] => ...
          )
  )
```

stephywells: Is this the output after it runs through the menu combine function?

shervElmi: Yes, I've tried possible conditions.

stephywells: It looks like you would loop through and check if [2] === 'wpbdp_admin'.

I'm seeing this if the extra menu isn't removed:
```
[5]=>
--
  | array(7) {
  | [0]=>
  | string(9) "Directory"
  | [1]=>
  | string(10) "edit_posts"
  | [2]=>
  | string(32) "edit.php?post_type=wpbdp_listing"
  | [3]=>
  | string(0) ""
  | [4]=>
  | string(46) "menu-top menu-icon-wpbdp_listing menu-top-last"
  | [5]=>
  | string(24) "menu-posts-wpbdp_listing"
  | [6]=> ...
  | }
```

This test needs to make sure 'edit.php?post_type=wpbdp_listing' is not included, and 'wpbdp_admin' is included.

shervElmi: Where did you get this output?

shervElmi:
@stephywells Please check the codes without tests that have the Todo comment above them, I postponed them.

garretlaxton (approved):
This is looking great!

The navigation to the left does look a bit weird for Authors and Contributors the way it's cut off. 
image

Would it be possible to extend that to fit the screen like it does for editor roles?
image

shervElmi:
> This is looking great!
> 
> The navigation to the left does look a bit weird for Authors and Contributors the way it's cut off. image
> 
> Would it be possible to extend that to fit the screen like it does for editor roles? image

@stephywells What do you think about these suggestions?

stephywells (approved):
Thanks @shervElmi !

PR #285, opened 2023-01-19
New Design: Base and top search bar and filtering styles

shervElmi:
With this Pull Request, we will redesign the top search bar and filtering, and prepare BD Core for new designs.

@stephywells I've restructured styles files, and I'm going to keep old codes and create an option in the settings for selecting new designs and old designs. I'll add this option after this PR.


stephywells:
@shervElmi was this closed intentionally?

PR #286, opened 2023-01-25
Fix listings don't appear correctly

shervElmi:
### Issue link:
https://github.com/Strategy11/business-directory-premium/issues/122

With this PR, we will fix the listing title that doesn't appear correctly in version `6.3.2`.


stephywells (changes requested):
I mentioned this in Slack, but adding it here to make it easier to keep track.

Let's move the default theme and grunt changes to a new PR. It looks like that may need more adjustment so we don't include the default theme style when the default theme isn't being used.

garretlaxton:
Are there any themes specifically this was happening on? I notice listing titles don't show in Enfold theme if the header title is hidden.

shervElmi:
> Are there any themes specifically this was happening on? I notice listing titles don't show in Enfold theme if the header title is hidden.

`X - The Theme`, `Enfold`, and WP default theme needs to test.

shervElmi:
@stephywells I tested all the themes and when I activate them, the default theme style isn't added to the page!
And for this PR to work properly, the default theme style must be here.

garretlaxton:
@shervElmi Do we have a copy of x theme? I don't see the listing titles in Enfold theme as mentioned.

shervElmi:
> @shervElmi Do we have a copy of x theme? I don't see the listing titles in Enfold theme as mentioned.

After your comment, I add some commits, and please clean your cache and test again because I can see that.

And you need to active the default theme.

shervElmi:
> @shervElmi Do we have a copy of x theme? I don't see the listing titles in Enfold theme as mentioned.

We already tested `X - The Theme`, and If you see it on the enfold theme, It'll be worked.

garretlaxton (approved):
Thanks! I see the titles now with the recent changes!
image

stephywells (changes requested):
It looks like this new css needs to be in the main css rather than only in the default theme, right? Is there a reason it wouldn't be needed in a different theme?
```
.wpbdp-listing .listing-title.show-listing-title {
	display: block;
}
```

shervElmi:
> It looks like this new css needs to be in the main css rather than only in the default theme, right? Is there a reason it wouldn't be needed in a different theme?
> 
> ```
> .wpbdp-listing .listing-title.show-listing-title {
> 	display: block;
> }
> ```

I saw `.wpbdp-listing .listing-title` in this file, and I decided to add my codes in the default theme CSS file.

stephywells:
@shervElmi It looks like this is part of the issue. We'll need to include that in the main css file to is works for an BD template. Can you move that?

shervElmi:
@stephywells I just tested all themes, and it works in the following themes:

- Business Card Theme
- Restaurant Theme
- Real Estate Directory Theme
- Mobile Compact Theme

And the template file of other themes is old, and we need to update them to make the new codes work.

stephywells (approved):
Thank you! This looks good to me now.

shervElmi:
@garretlaxton Would you please check all themes and open a GH issue for the theme that doesn't show the listing title?

PR #289, opened 2023-02-14
New Design - Improve Local Environment Tools

shervElmi:
With this PR, we will improve the plugin development tools.

### Less live-reload
After saving any style files, my site fetches new styles without refreshing.
We need to run the `grunt watch` command and add the following codes with the Code Snippets plugin on our site.

```
if ( !is_admin() && in_array($_SERVER['REMOTE_ADDR'], array('127.0.0.1', '::1'))) {
  wp_register_script('livereload', 'http://localhost:35729/livereload.js?snipver=1', null, false, true);
  wp_enqueue_script('livereload');
}
```

### Auto-Reorder and Format Style files
As you can see in the following Gif, after I saved my file, the editor automatically reordered and formatted my file.

It also works with the `npm run format:css` command.

### All Available Commands:

```
npm run format
npm run format:css
npm run format:js
npm run format:php

npm run lint
npm run lint:css
npm run lint:js
npm run lint:php
```


stephywells (commented):
This looks cool! Will this be a pain to maintain going forward?

shervElmi:
@stephywells I don't think so. Since we use Grunt, it should work fine.

stephywells:
@shervElmi how much maintenance will this need for the css ordering?

shervElmi:
@stephywells There won't be any maintenance required, as it's just some Stylelint rules.

stephywells (approved):
Thanks for the extra detail! Before merging this, can you add the code you included in the PR description somewhere in the plugin where it can be uncommented or something? We'll need it documented in the code.

shervElmi:
@stephywells I have received this error today: PHPStan Code Analysis: All jobs have failed!

includes/class-assets.php
```diff
@@ -180,6 +180,12 @@ public function enqueue_scripts() {
 				WPBDP_VERSION
 			);
 		}
+
+		// Enable `grunt-contrib-watch` livereload.
+		// Live reload server will be started with the watch task per target.
+		if ( defined( 'SCRIPT_DEBUG' ) && SCRIPT_DEBUG && in_array( $_SERVER['REMOTE_ADDR'], [ '127.0.0.1', '::1' ] ) ) {
+			wp_enqueue_script( 'livereload', 'http://localhost:35729/livereload.js?snipver=1', [], WPBDP_VERSION, true );
```
stephywells, line 187: Can we use the site url for this? And where does the livereload.js come from? Is it created with an npm command?

stephywells (changes requested):

stephywells:
> I received this error `PHPStan Code Analysis: All jobs have failed!` today for every commit!

It looks like it's running, but using a new "is unused" check. I think for right now, we can go ahead and exclude that check in the phpstan config.

includes/class-assets.php
shervElmi: No! It generates by Grunt in [this line](https://github.com/Strategy11/business-directory-plugin/pull/289/files#diff-8e7e5f941c67b0cf481a9d650295119d4922b96117e7adc069ba30a81a78c65fR90) when we run the `grunt watch` command with this fixed URL.

stephywells: Oooh!! This is really cool! I didn't test it before, and I'm not sure why that url works, but it's magical.

stephywells (approved):

PR #293, opened 2023-02-17
fix Rankmath conflict

stephywells:
I'm not sure why this was causing an issue, but I'm not seeing a reason to return $content by reference.

Fixes https://github.com/Strategy11/business-directory-premium/issues/120

@shervElmi do you see any possible issues with this?


shervElmi (approved):
It doesn't change the $content variable, but of course we don't need the content variable with a reference. Also, I've checked all the plugins that use this function, and there is no risk. Thank you so much!

stephywells:
Thanks so much for double checking this @shervElmi !

PR #301, opened 2023-04-13
Fix: Correct always-false if condition in class-csv-import.php PHPStan

shervElmi:
This PR addresses an issue in the `includes/admin/helpers/csv/class-csv-import.php` file where an if condition was always evaluating to false during PHPStan checks.

### DIFF
--- a/includes/admin/helpers/csv/class-csv-import.php
+++ b/includes/admin/helpers/csv/class-csv-import.php
@@ -653,7 +653,7 @@ class WPBDP_CSV_Import {
 			}
 		}
 
-		if ( $state->images ) {
+		if ( ! empty( $state->images[0] ) ) {
 			$listing->set_thumbnail_id( $state->images[0] );
 		}
 


includes/admin/helpers/csv/class-csv-import.php
```diff
@@ -653,7 +653,7 @@ private function import_row( $data ) {
 			}
 		}
 
-		if ( $state->images ) {
+		if ( is_array( $state->images ) && ! empty( $state->images ) && ! empty( $state->images[0] ) ) {
```
stephywells, line 656: Thank you @shervElmi ! Does this pass without "is_array( $state->images )" too?

shervElmi: You're welcome, I have made the necessary changes in my recent commit, and it now passes without using "is_array( $state->images )".

stephywells (approved):

includes/admin/helpers/csv/class-csv-import.php
stephywells: Thanks! That's nice to have it a bit shorter now

PR #318, opened 2023-05-02
fix sidebar list titles

mostafa-hisham:
Fixes #[145](https://github.com/Strategy11/business-directory-premium/issues/154) (premium)

### DIFF
--- a/includes/class-wordpress-template-integration.php
+++ b/includes/class-wordpress-template-integration.php
@@ -129,7 +129,7 @@ class WPBDP__WordPress_Template_Integration {
 	 * @return string
 	 */
 	public function set_tax_title( $title ) {
-		if ( $this->in_the_loop() ) {
+		if ( ! $this->in_the_loop() ) {
 			// If this is not a category, don't change it.
 			return $title;
 		}
--- a/includes/class-wordpress-template-integration.php
+++ b/includes/class-wordpress-template-integration.php
@@ -195,7 +195,7 @@ class WPBDP__WordPress_Template_Integration {
 	 * @return string
 	 */
 	public function remove_tax_thumbnail( $thumbnail ) {
-		if (  $this->in_the_loop() ) {
+		if ( $this->in_the_loop() ) {
 			return $thumbnail;
 		}
 		// The caption shows in 2021 theme.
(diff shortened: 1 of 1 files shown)


stephywells:
@garretlaxton would you mind looking over this one?

stephywells:
@mostafa-hisham I haven't tested this out yet, but just looked over the code. Let me lay out a few different scenarios that this affects, just so we can be sure you've checked them since this is a complicated issue.
- When viewing a category, the page should have the category title, and each listing should have the correct listing title (This works a bit differently for block themes and non block themes)
- There is a setting called 'Disable advanced CPT integration'. I think this setting was originally added as a workaround after switching the listings to a CPT. I'd like to get rid of this setting, but for now we need to check this too.
- When viewing the directory, every listing should have the correct title, and the directory page should not use a title from a listing
- If a listing has a thumbnail, it should only be used for that listing, and not applied to the whole page

I'm not sure if this is helpful at all, but just wanted to give a bit more detail.

garretlaxton:
This fixes the issue, but it also brings back https://github.com/Strategy11/business-directory-premium/issues/141 that Steph mentioned.

mostafa-hisham:
screenshoots form category page (cat1) `/business-directory/wpbdp_category/cat1/`
using Wordpress v6.2

enfold
astra

mostafa-hisham:
@garretlaxton could you please check it again and tell me which WordPress version you are using?

garretlaxton (approved):
Nice, this is looking good now! I am using 6.2 btw.

stephywells (changes requested):
@mostafa-hisham It looks like there are a lot of changes here that are against our code styling rules. This whole plugin doesn't yet use tabs, but we are moving toward them.

Also, I'm seeing spaces removed in lines like line 15:
add_action( 'body_class', array( $this, 'add_basic_body_classes' ) );

Can you please revert any codestyling changes to make it clear what is changing? We like to submit any codestyling changes in its own PR.

stephywells (approved):
Thanks for all your work and testing on this @mostafa-hisham and @garretlaxton !

The code looks great. I haven't tested it, but it sounds like we can go ahead and release it.

PR #322, opened 2023-05-19
Pass new var to listing template to check if it has excerpt field

mostafa-hisham:
Will help fix #[6](https://github.com/Strategy11/bd-business-card/issues/6) and #[5](https://github.com/Strategy11/bd-business-card/issues/5)

### DIFF
--- a/includes/helpers/class-listing-display-helper.php
+++ b/includes/helpers/class-listing-display-helper.php
@@ -96,7 +96,7 @@ class WPBDP_Listing_Display_Helper {
 			}
 			$vars['listing_css_class'] .= ' with-image';
 		}
-
+		$vars['listing_has_excerpt'] = $vars['fields']->has_excerpt();
 		$vars = apply_filters( 'wpbdp_listing_template_vars', $vars, $post_id );
 		$vars = apply_filters( 'wpbdp_single_template_vars', $vars, $post_id );
 
--- a/includes/helpers/class-field-display-list.php
+++ b/includes/helpers/class-field-display-list.php
@@ -11,8 +11,7 @@ class WPBDP_Field_Display_List implements IteratorAggregate {
 	private $items            = array();
 	private $displayed_fields = array();
 	private $names_to_ids     = array();
-
-
+	private $has_excerpt     = false;
 	public function __construct( $listing_id, $display, $fields = array() ) {
 		$this->listing_id = $listing_id;
 		$this->display    = $display;
@@ -52,6 +51,10 @@ class WPBDP_Field_Display_List implements IteratorAggregate {
 		// if( $f->display_in( $this->display ) )
 		//     $this->displayed_fields[] = $field_id;
 
+		if ( $f->get_tag() === 'excerpt' ) {
+			$this->has_excerpt = true;
+		}
+
 		$this->displayed_fields[]                   = $field_id;
 		$this->items[ $field_id ]                   = new _WPBDP_Lightweight_Field_Display_Item( $f, $this->listing_id, $this->display );
 		$this->names_to_ids[ $f->get_short_name() ] = $field_id;
@@ -60,6 +63,10 @@ class WPBDP_Field_Display_List implements IteratorAggregate {
 
 	public function freeze() {
 		$this->frozen = true;
+	}
+
+	public function has_excerpt() {
+		return $this->has_excerpt;
 	}
 
 	public function not( $filter ) {


includes/helpers/class-listing-display-helper.php
```diff
@@ -96,7 +96,7 @@ public static function single_listing_vars( $include = array() ) {
 			}
 			$vars['listing_css_class'] .= ' with-image';
 		}
-
+		$vars['listing_has_excerpt'] = $vars['fields']->has_excerpt();
```
stephywells, line 99: Can we handle this in the WPBDP_Field_Display_List class too? Everything else is returned conditionally, so it looks like this may result in an unexpected response. Would $vars['fields'] always be set before this point?

stephywells (changes requested):
Nice job finding the code for this!

.github/workflows/phpcs.yml
```diff
@@ -21,7 +21,7 @@ jobs:
           export PHPCS_DIR=/tmp/phpcs
           export SNIFFS_DIR=/tmp/sniffs
           git clone -b master --depth 1 https://github.com/squizlabs/PHP_CodeSniffer.git $PHPCS_DIR
-          git clone -b master --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git $SNIFFS_DIR
+          git clone -b 2.3.0 --depth 1 https://github.com/WordPress-Coding-Standards/WordPress-Coding-Standards.git $SNIFFS_DIR
```
stephywells, line 24: Looks like they changed 'master' to 'main'. This would be easier to maintain if we can avoid hardcoding a version number.

includes/themes.php
```diff
@@ -964,6 +969,18 @@ public function install_theme( $file ) {
...
+	/**
+	 * Shows admin notice to update the plugin.
+	 *
+	 * @returns {void}
+	 */
+	public function update_bd_admin_notice() {
+		$class = 'notice notice-warning';
+		$message = __( 'Please update business directory plugin to latest version.', 'business-directory-plugin' );
+
+		printf( '<div class="%1$s"><p>%2$s</p></div>', esc_attr( $class ), esc_html( $message ) );
```
stephywells, line 982: Instead of printing the notice directly, let's use the message functions that already exist. (wpbdp_admin_message)

includes/themes.php
```diff
@@ -43,6 +43,11 @@ function __construct() {
 			require_once WPBDP_PATH . 'includes/admin/controllers/class-themes-admin.php';
 			$this->admin = new WPBDP_Themes_Admin( $this, wpbdp()->licensing );
 		}
+
+		$theme_requires = $this->get_active_theme_data( 'requires' );
```
stephywells, line 47: This function is already quite long. Can we move this new code to a new function?

includes/helpers/class-listing-display-helper.php
stephywells: Looks great!

stephywells (changes requested):
Thanks for adding the requires check @mostafa-hisham ! That will be really nice for the future.

.github/workflows/phpcs.yml
stephywells: Thanks! Can you please fix that error so we don't get stuck on v2.x? We can fix it in a different PR.

includes/themes.php
```diff
@@ -964,6 +966,26 @@ public function install_theme( $file ) {
 	public function sync_settings() {
 		_deprecated_function( __METHOD__, '5.0' );
 	}
+
+	/**
+	 * Check current theme requires plugin version and show admin notice if it doesn't meet the requirement  .
+	 *
+	 * @returns {void}
+	 */
+	private function check_theme_requiers_version() {
```
stephywells, line 975: Let's fix the typo in this function name.

includes/themes.php
```diff
@@ -964,6 +966,26 @@ public function install_theme( $file ) {
...
+		$theme_requires = $this->get_active_theme_data( 'requires' );
+		if ( defined( 'WPBDP_VERSION' ) && $theme_requires && version_compare( WPBDP_VERSION, $theme_requires, '<' ) ) {
+			add_action( 'admin_notices', array( $this,'update_bd_admin_notice' ) );
```
stephywells, line 978: Does this work if we add is_admin() to the checks and move the new wpbdp_admin_message line here instead of using a hook? It would be nice if we didn't have to add a new public function if we don't have to since private functions are easier to maintain and remove.

includes/themes.php
```diff
@@ -332,7 +334,7 @@ function get_active_theme() {
 	/**
 	 * Retrieves theme information for the current active theme.
 	 *
-	 * @return object
+	 * @return object|string|boolean
```
stephywells, line 337: This change seems to be breaking a log of PHPstan rules. I wonder if we need to add a new function without $key so we know an object will always be returned? Something like this?

```
public function get_active_theme_data( $key = null ) {
		$data   = $this->get_active_theme_object();
		if ( ! is_null( $key ) ) {
			return isset( $data->{$key} ) ? $data->{$key} : false;
		}
		return $data;
	}

public function get_active_theme_object() {
		$active = $this->get_active_theme();
		$data   = $this->themes[ $active ];
		return $data;
	}
```

stephywells (changes requested):

includes/themes.php
stephywells: We can use leave the hook then. Thanks for checking!

.github/workflows/phpcs.yml
stephywells: For this PR, we can switch it back to 2.x like you had it. Then we can still get it merged and not hold up other PRs.

includes/themes.php
```diff
@@ -964,6 +966,26 @@ public function install_theme( $file ) {
...
+			add_action( 'admin_notices', array( $this,'update_bd_admin_notice' ) );
+		}
+	}
+	/**
+	 * Shows admin notice to update the plugin.
+	 *
+	 * @returns {void}
+	 */
+	public function update_bd_admin_notice() {
+		wpbdp_admin_message( __( 'Please update business directory plugin to latest version.', 'business-directory-plugin' ) );
```
stephywells, line 987: Let's use the plugin name in caps:
```suggestion
		wpbdp_admin_message( __( 'Please update Business Directory Plugin to latest version.', 'business-directory-plugin' ) );
```

includes/themes.php
```diff
@@ -964,6 +966,26 @@ public function install_theme( $file ) {
 	public function sync_settings() {
 		_deprecated_function( __METHOD__, '5.0' );
 	}
+
+	/**
+	 * Check current theme requires plugin version and show admin notice if it doesn't meet the requirement  .
+	 *
+	 * @returns {void}
+	 */
+	private function check_theme_requiered_version() {
```
stephywells, line 975:
```suggestion
	private function check_theme_required_version() {
```
Can we update the typo here?

includes/themes.php
```diff
@@ -43,6 +43,8 @@ function __construct() {
 			require_once WPBDP_PATH . 'includes/admin/controllers/class-themes-admin.php';
 			$this->admin = new WPBDP_Themes_Admin( $this, wpbdp()->licensing );
 		}
+
+		$this->check_theme_requiered_version();
```
stephywells, line 47:
```suggestion
		$this->check_theme_required_version();
```
Just a typo here too.

stephywells (changes requested):
I made a few suggestions. That should wrap up this PR.

Thanks for the great solution!

stephywells (approved):
Yay! Looks perfect. Thanks so much!

PR #328, opened 2023-06-22
fix upload image after delete

mostafa-hisham:
fix #[160](https://github.com/Strategy11/business-directory-premium/issues/160)


mostafa-hisham:
@garretlaxton could you please take another look at it? I fixed the issue

garretlaxton (approved):
Nice, this looks great now!

assets/js/wpbdp.js
```diff
@@ -275,6 +275,9 @@ WPBDP.fileUpload = {
                         if ( ! t._admin_nonce ) {
                             t._slotsRemaining++;
                             $( '#image-slots-remaining' ).text( t._slotsRemaining );
+							if(t._slotsRemaining < t._slots){
```
stephywells, line 278: Let's update this line to match the codestyling in other lines. It should be:
`if ( t._slotsRemaining < t._slots ) {`

stephywells (changes requested):
This works great! Let's fix the codestyling here too, and then we can marge it.

stephywells (approved):
Perfect!

PR #329, opened 2023-06-23
Fix phpstan errors

mostafa-hisham:
fix [php stan errors](https://github.com/Strategy11/business-directory-plugin/actions/runs/5348143233/jobs/9697662306?pr=328)

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -10,6 +10,7 @@ parameters:
 		- */vendors/*
 		- */languages/*
 		- */tests/*
+		- */node_modules/*
 		- stubs.php
 	ignoreErrors:
 		- '#Access to an undefined property#'
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -12,6 +12,7 @@ parameters:
 		- */tests/*
 		- stubs.php
 		- */node_modules/*
+		- includes/compatibility/deprecated/*
 	ignoreErrors:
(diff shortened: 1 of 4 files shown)


phpstan.neon
```diff
@@ -19,3 +22,5 @@ parameters:
 		- '#Method PclZip::extract\(\) invoked with #'
 		- '#overridden property WP_List_Table::\$items#'
 		- '#Method WPBDP_[a-zA-Z0-9\\_]+::[a-zA-Z0-9\\_]+\(\) is unused#'
+		- '#Call to function is_wp_error\(\) with array will always evaluate to false#'
+		- '#function wp_update_post expects array#'
```
stephywells, line 26: Can we fix this error instead of excluding them?

It looks like the wp_update_post line causing the error should be:
```
wp_update_post(
array(
'ID' => $img->ID, 
'post_excerpt' => $value[1],
)
);
```

For the is_wp_error error, I'm not sure why it's having trouble on that one row. Let's exclude the error in that one spot instead of allowing it globally.
https://phpstan.org/user-guide/ignoring-errors#ignoring-in-code-using-phpdocs

stephywells (changes requested):
Thanks for fixing these!

Let's adjust this just a bit so we ignore as few rules as possible.

stephywells (approved):
This looks great! Thanks for the updates!

PR #335, opened 2023-08-07
convert Null values of is_sticky to 0 to avoid sort problems

mostafa-hisham:
should help fixes #[33](https://github.com/Strategy11/business-directory-ratings/issues/33) when there is sticky listing


garretlaxton (approved):
This looks good as well!

stephywells (approved):
Looks good!

Just curious @mostafa-hisham . Have you checked the query speed to see if this affects it?

stephywells:
@mostafa-hisham Thanks for checking! Is there a better way of doing this that won't add extra load on the DB? What happens if we run the SELECT sticky query separately?

stephywells:
@mostafa-hisham I was thinking something like:
`$sticky_posts = "SELECT listing_id FROM {$wpdb->prefix}wpbdp_listings WHERE is_sticky=1";`

And then in the main query, we would add this to the WHERE query:
```
if ( $sticky_posts ) {
   ' AND in (' . implode( $sticky_posts, ',' ) . ')';
}
```

I think this might cut down the time, but I'm not positive.

stephywells:
Oh, right. But I think the same logic would apply, but using the ids for sorting. Right?

includes/class-query-integration.php
```diff
@@ -249,13 +249,11 @@ public function posts_clauses( $pieces, $query ) {
 		$pieces = apply_filters( 'wpbdp_query_clauses', $pieces, $query );
 
 		// Sticky listings.
-		$is_sticky_query = "(SELECT is_sticky FROM {$wpdb->prefix}wpbdp_listings wls WHERE wls.listing_id = {$wpdb->posts}.ID LIMIT 1) AS wpbdp_is_sticky";
-
+		$sticky_ids_str = $this->get_sticky_listing_ids();
 		if ( in_array( wpbdp_current_view(), wpbdp_get_option( 'prevent-sticky-on-directory-view' ), true ) ) {
-			$is_sticky_query = '';
+			$sticky_ids_str = '';
```
stephywells, line 254: Can we reverse the logic here, so we only run the DB query when we use it?

includes/class-query-integration.php
```diff
@@ -282,12 +280,26 @@ public function posts_clauses( $pieces, $query ) {
...
 	}
 
+	/**
+	 * Select all sticky listing ids and return thim in string comma seperated.
+	 *
+	 * @return string listing ids
+	 */
+	private function get_sticky_listing_ids() {
+		global $wpdb;
+		$results = $wpdb->get_results( "SELECT listing_id FROM {$wpdb->prefix}wpbdp_listings  WHERE is_sticky=1" );
```
stephywells, line 295: Let's run the query through our check_cache() functions too.

stephywells (changes requested):
Thanks for trying that! I haven't tested it, but the code looks good, and that's great that it speeds it up a bit.

includes/class-query-integration.php
stephywells: Thanks for the update!

The group here should be for the table getting searched. This helps us keep track of when the cache should be cleared.

`'group' => 'wpbdp_listings',`

stephywells (changes requested):
Just left one more comment. This could use a final round of testing too since we made some updates.

includes/class-query-integration.php
stephywells: I made this change and merged in my changes from the other pr.

stephywells (approved):

PR #340, opened 2023-09-18
Fix issue with custom db prefix

stephywells:
Fixes https://github.com/Strategy11/business-directory-premium/issues/174


garretlaxton (approved):
This works great! I also made a quick test site with a different DB prefix which works as normal now.

PR #350, opened 2023-11-16
Fix https://github.com/Strategy11/business-directory-premium/issues/188

stephywells:
loop_start filter wasn't firing when the category template is switched in WP 6.4

To see the issue, activate a block theme and update to WP 6.4.1. The listings don't show on category pages.

I've tested this fix with 2023 theme in WP 6.4.1 and WP 6.3.2. It could use more testing in more themes though.

### DIFF
--- a/includes/class-wordpress-template-integration.php
+++ b/includes/class-wordpress-template-integration.php
@@ -116,7 +116,7 @@ class WPBDP__WordPress_Template_Integration {
 	 * @return void
 	 */
 	public function setup_tax_hooks() {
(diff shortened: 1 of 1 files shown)


garretlaxton:
I'm seeing the same issue for region pages as well.

stephywells:
Oh, good catch! That's updated now.

garretlaxton (approved):
I tested quite a bit of themes, this looks good!

stephywells:
Awesome! Thanks so much!!

PR #369, opened 2024-02-27
Fix possible undefined error in search widget

stephywells:
Only use advanced search if fields are selected
Set defaults to avoid errors
Fixes https://github.com/Strategy11/business-directory-premium/issues/208


garretlaxton (approved):
This looks good!

PR #376, opened 2024-05-01
Cut down more api calls

stephywells:
Prepare for release


stephywells:
There are a lot of good translation issues recommended here from Code Rabbit. But I don't have the time to implement them and our changes may not be used in WP anyway.

PR #383, opened 2024-06-20
Fix reported Fatal Error (#236)

banago:
`get_latest_payment()` can return null, but was not checked for that, so it caused a fatal error in the scenario when it is not an object.

### DIFF
--- a/includes/controllers/pages/class-manage-listings.php
+++ b/includes/controllers/pages/class-manage-listings.php
@@ -107,7 +107,7 @@ class WPBDP__Views__Manage_Listings extends WPBDP__View {
 		$is_pending_payment = ( 'pending_payment' === $listing_status );
 		$payment_url        = $listing->get_payment_url();
 
-		if ( $payment_url !== null ) {
+		if ( $payment_url ) {
 			$buttons = sprintf(
 				'<a class="button wpbdp-button renew-listing" href="%s" %s >%s</a>',
 				$is_pending_payment ? esc_url( $payment_url ) : esc_url( $listing->get_renewal_url() ),
--- a/includes/controllers/pages/class-manage-listings.php
+++ b/includes/controllers/pages/class-manage-listings.php
@@ -108,13 +108,15 @@ class WPBDP__Views__Manage_Listings extends WPBDP__View {
 		$payment_url        = $listing->get_payment_url();
 		$show_button        = ! $is_pending_payment || $payment_url;
 
-		if ( $show_button ) {
-			$buttons = sprintf(
-				'<a class="button wpbdp-button renew-listing" href="%s" target="_blank" rel="noopener">%s</a>',
-				$is_pending_payment ? esc_url( $payment_url ) : esc_url( $listing->get_renewal_url() ),
-				$is_pending_payment ? esc_html__( 'Pay Now', 'business-directory-plugin' ) : esc_html__( 'Renew Listing', 'business-directory-plugin' )
-			) . $buttons;
+		if ( ! $show_button ) {
+			return $buttons;
 		}
+
+		$buttons = sprintf(
+			'<a class="button wpbdp-button renew-listing" href="%s" target="_blank" rel="noopener">%s</a>',
+			$is_pending_payment ? esc_url( $payment_url ) : esc_url( $listing->get_renewal_url() ),
+			$is_pending_payment ? esc_html__( 'Pay Now', 'business-directory-plugin' ) : esc_html__( 'Renew Listing', 'business-directory-plugin' )
+		) . $buttons;
 
 		return $buttons;
 	}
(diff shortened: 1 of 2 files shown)


includes/controllers/pages/class-manage-listings.php
```diff
@@ -105,12 +105,16 @@ public function maybe_add_renew_button( $buttons, $listing_id ) {
...
-			$is_pending_payment ? esc_url( $listing->get_payment_url() ) : esc_url( $listing->get_renewal_url() ),
-			'target="_blank" rel="noopener"',
-			$is_pending_payment ? esc_html__( 'Pay Now', 'business-directory-plugin' ) : esc_html__( 'Renew Listing', 'business-directory-plugin' )
-		) . $buttons;
+		$payment_url        = $listing->get_payment_url();
+
+		if ( $payment_url ) {
+			$buttons = sprintf(
+				'<a class="button wpbdp-button renew-listing" href="%s" %s >%s</a>',
+				$is_pending_payment ? esc_url( $payment_url ) : esc_url( $listing->get_renewal_url() ),
```
stephywells, line 113: Since we are checking for the payment url, are there cases where the renewal url will be missing when it should be included when the payment is not pending?

stephywells (commented):
Thanks for the PR! I just have a question.

includes/controllers/pages/class-manage-listings.php
banago: I did consider this and I assumed the renewal function was also dependent on the Payment object and would faill 100% of the time if no payment was present. 

I checked the renewal function again and tested it's internals. You are right, it uses the $listing object and works regardless. 

I've refactored the conditionals to fix this.

includes/models/class-listing.php
```diff
@@ -621,7 +621,7 @@ public function get_renewal_url( $deprecated = 0 ) {
 	 */
 	public function get_payment_url() {
 		$payment = $this->get_latest_payment();
-		return $payment->get_checkout_url();
+		return is_object( $payment ) && method_exists( $payment, 'get_checkout_url' ) ? $payment->get_checkout_url() : null;
```
stephywells, line 624: The return type for this function is a string. Can we return '' instead of null to keep the same return type?

includes/controllers/pages/class-manage-listings.php
stephywells: This helps, but I don't think we want to show the button if pending and no url. Falling back to the renewal url doesn't really make sense to me if the payment is still pending.

stephywells (changes requested):

includes/controllers/pages/class-manage-listings.php
banago: This isn't right, I pushed different code. Checking.

banago: This is outdated. Please review latest changes.

includes/models/class-listing.php
banago: Done.

includes/controllers/pages/class-manage-listings.php
stephywells: I'm not seeing new changes for this. I think we'll need to add a check like this:
`$show_button = !$is_pending_payment || $payment_url;`

includes/models/class-listing.php
stephywells: Thank you!

includes/controllers/pages/class-manage-listings.php
banago: That's the approach I took. Please review the code here:

https://github.com/Strategy11/business-directory-plugin/blob/087b91854d6af9d613457a2b8d4d198b1e4778ff/includes/controllers/pages/class-manage-listings.php#L110

stephywells: @banago That's not quite the same thing. Your version will show the Renew listing button, even if the payment is pending. It would better not to show the button if the payment is pending and there's no payment url.

banago: @stephywells got it, thanks.

banago:
@stephywells I pushed an update.

includes/controllers/pages/class-manage-listings.php
stephywells: Can we do this without adding duplication? Here's what I was thinking:

```suggestion
		
		$is_pending_payment = ( 'pending_payment' === $listing_status );
		$pending_url                 = $listing->get_payment_url();
		$show_button               = ! $is_pending_payment || $payment_url;
		if ( $show_button ) {
		    $buttons            = sprintf(
			    '<a class="button wpbdp-button renew-listing" href="%s" %s >%s</a>',
			    $is_pending_payment ? esc_url( $pending_url ) : esc_url( $listing->get_renewal_url() ),
			    'target="_blank" rel="noopener"',
			    $is_pending_payment ? esc_html__( 'Pay Now', 'business-directory-plugin' ) : esc_html__( 'Renew Listing', 'business-directory-plugin' )
		    ) . $buttons;
		}
```

stephywells (approved):
Thank you for the update! This looks much safer now.

Fixes https://github.com/Strategy11/business-directory-premium/issues/236

PR #243, opened 2022-08-18
Use separate setting for main thumbnail size

stephywells:
Fixes https://github.com/Strategy11/BusinessDirectoryPlugin/issues/5130

This allows for the two different settings sizes to be used.

This rearranges the thumbnail settings and renames a couple settings:
- gallery thumbnail width/height
- Change "Directory thumb" in dropdown to "Default (size set below)"
- change "default thumbnail size" to "Main thumbnail inage size"

PR #384, opened 2024-06-20
Adds Tiktok as a social media option. Fixes #243

banago:
fixes https://github.com/Strategy11/business-directory-premium/issues/243


assets/images/social/Tiktok.svg
```diff
@@ -0,0 +1,12 @@
+<svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" version="1.1"
```
stephywells, line 1: Looks like we can minify this to save 50%. I use this:
https://jakearchibald.github.io/svgomg/

stephywells (changes requested):
This looks great! Thanks for fixing Reddit too.

assets/images/social/Tiktok.svg
banago: Good call. I did not bother initially in case we would change that for another variation of the logo.

sorinmarta (approved):

stephywells (approved):

PR #385, opened 2024-06-21
Hides Expiration and Renewal options from settings

banago:
If "listing-renewal" option is set to OFF, three options from Email Notifications settings are hidden through CSS.

Fixes https://github.com/Strategy11/business-directory-premium/issues/238


banago:
@stephywells someone will need to run the `grunt` commands on this CSS change. I could not get it working because I could not deduce all the required packages to install.

includes/class-assets.php
```diff
@@ -525,7 +525,14 @@ public function enqueue_admin_scripts( $force = false ) {
 	 */
 	public function add_body_class( $admin_body_classes ) {
 		if ( WPBDP_App_Helper::is_bd_page() ) {
-			$admin_body_classes = ' wpbdp-admin-page';
+			$admin_body_classes .= ' wpbdp-admin-page';
+
+			// Hide expiration/renewal options, when renewal is OFF.
+			if( ! wpbdp_get_option( 'listing-renewal' ) ){
+				$admin_body_classes .= ' no-renewal';
```
stephywells, line 532: This class isn't being used in the css. The added css is causing the options to always be hidden.

stephywells (changes requested):
This isn't working quite right for me. The settings are always hidden with this css.

banago:
@stephywells, just pushed the fix for this.

includes/class-assets.php
stephywells: Thanks for the fix! It would also be great to include our prefix on the class.
`wpbdp-no-renewal`

stephywells (changes requested):

assets/css/less/admin-settings.less
```diff
@@ -258,6 +258,12 @@
     }
 }
 
+/* Using the ID selector is necessary because there is no class-based identifier for this settings */
+.wpbdp-no-renewal #wpbdp-settings-admin-notifications .wpbdp-settings-multicheck-option-no-2,
+.wpbdp-no-renewal #wpbdp-settings-admin-notifications .wpbdp-settings-multicheck-option-no-3,
+.wpbdp-no-renewal #wpbdp-settings-user-notifications .wpbdp-settings-multicheck-option-no-3 {
```
stephywells, line 264: Let's take advantage of the Less format here to avoid repeating the .wpbdp-no-renewal in each line.

assets/css/less/admin-settings.less
```diff
@@ -258,6 +258,12 @@
     }
 }
 
+/* Using the ID selector is necessary because there is no class-based identifier for this settings */
+.wpbdp-no-renewal #wpbdp-settings-admin-notifications .wpbdp-settings-multicheck-option-no-2,
```
stephywells, line 262: It's possible that this generic class on the checkboxes could cause issues in the future if we add or rearrange options. Maybe we could add another class in the setting_multicheck_callback() function? It could include the $option_value so we know we are hiding the right options.

stephywells (changes requested):
Thank you! This is almost there. Let's just get it cleaned up a bit for release and easier maintainability.

assets/css/less/admin-settings.less
banago: That's a good idea. On it.

sorinmarta (approved):

stephywells (approved):

PR #386, opened 2024-06-25
Converts http to https

banago:
WordPress [appends http ](https://github.com/WordPress/wordpress-develop/blob/e8fed338f8f27363cb24f0ea2792cfa3b311b612/src/wp-includes/formatting.php#L4515)to santized urls. This converts that to https when we save our URL fields for sites that use SSL.

Fixes https://github.com/Strategy11/business-directory-premium/issues/228

### DIFF
--- a/includes/fields/class-fieldtypes-url.php
+++ b/includes/fields/class-fieldtypes-url.php
@@ -141,7 +141,8 @@ class WPBDP_FieldTypes_URL extends WPBDP_Form_Field_Type {
 		}
 
 		// Adds the http protocol if it's missing.
-		$url  = esc_url_raw( is_array( $input ) ? trim( $input[0] ) : $input );
+		$url_raw = esc_url_raw( is_array( $input ) ? trim( $input[0] ) : $input );
+		$url = is_ssl() ? str_replace( 'http://', 'https://', $url_raw ) : $url_raw;
 		$text = trim( is_array( $input ) ? sanitize_text_field( $input[1] ) : '' );
 
 		return array( $url, $text );
--- a/includes/fields/class-fieldtypes-url.php
+++ b/includes/fields/class-fieldtypes-url.php
@@ -140,7 +140,7 @@ class WPBDP_FieldTypes_URL extends WPBDP_Form_Field_Type {
 			return array( '', '' );
 		}
 
-		// Adds the http protocol if it's missing.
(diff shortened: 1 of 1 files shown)


includes/fields/class-fieldtypes-url.php
```diff
@@ -141,8 +141,9 @@ public function convert_input( &$field, $input ) {
 		}
 
 		// Adds the http protocol if it's missing.
-		$url  = esc_url_raw( is_array( $input ) ? trim( $input[0] ) : $input );
-		$text = trim( is_array( $input ) ? sanitize_text_field( $input[1] ) : '' );
+		$url_raw = esc_url_raw( is_array( $input ) ? trim( $input[0] ) : $input );
+		$url     = is_ssl() ? str_replace( 'http://', 'https://', $url_raw ) : $url_raw;
```
stephywells, line 145: Since this is for url values saved into website fields in a listing, I'm not sure that is_ssl has any bearing here. I'm not sure that this is a change we should force.

stephywells:
Adding a note here to link to the related issue. (We should be sure to include the link in the PR description)
https://github.com/Strategy11/business-directory-premium/issues/228

includes/fields/class-fieldtypes-url.php
banago: The `is_ssl()` works correctly because I tested it on a `http` and `https` environment respectively.

This fixes the issue where a schema-less url will be appended `http://` on a `https` site and saved to database.

stephywells: The links we are changing here are external links that are added by website users adding or editing listings. I don't think we can blanket switch all of these types of urls to https.

banago: We are forcing them to go with `http` though. It's better to err on the side of `https`.

If the URL has another schema, such as `tel:`, `mailto:`, etc., this won't take effect. It only applies to schema-less URLs.

stephywells: esc_url_raw only adds http if it was not already included in the url. This update will always force the url to https, even if you try to switch it to http.

banago: That's right, this overwrites all URLs that are saved, even those that are intentionally entered as `http://`. I overlooked this.

PR #392, opened 2024-09-07
Add PHPStorm config

sorinmarta:
This PR adds the starter config for PHPStorm and then it adds the folder to `.gitignore` so that it will prevent people from adding their custom configs by mistake.


sorinmarta:
@stephywells 

While waiting to get access to Helpscout I started with adding the PHPStorm configs so that when I'll get access I can make new branches with the config in them.

I think this will be a very repetitive task for you to go over these PRs for all of the WPT, BD, and AWP plugins. 

Perhaps we can agree that for these PRs I'll merge them without review since the files and changes will be exactly the same every time and they will not influence the releases in any way. I'll also make sure they are excluded in buddy.works every time I merge a PR.

This way you won't have to always go through and accept the PRs for over 30 repos 😄 

I'll leave them as waiting for review for this repo and the premium one. If you agree with my idea, we need to change the repo settings to allow me to merge without a review so that I can start merging the other repos.

Please let me know your thoughts on that!

stephywells (changes requested):
Thank you!

Looks like this one needs to be excluded in .gitattributes too.

sorinmarta:
>Looks like this one needs to be excluded in .gitattributes too

Done, thanks!

>Maybe you could do a search in each product for ".git" or something to make sure the new folder gets excluded everywhere it needs to be?

Indeed, I'll check all the other repos for `.gitattributes`, `.distignore`, and similar files.

Thanks!

stephywells (approved):
Thank you!

PR #393, opened 2024-09-10
Fix memory leak & infinite loop

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/248

This PR tweaks a small memory leak which was causing an infinite loop when including the template files.


sorinmarta:
This change is no longer needed.

PR #394, opened 2024-09-11
Add HTTPS if no protocol is used in the URL Field

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/228

This PR checks if the value entered in a URL Field has either HTTP or HTTPS. If it doesn't have it, it defaults to HTTPS.


stephywells (approved):
Great solution! Very nice!

PR #395, opened 2024-09-12
Add empty check for is_sticky property

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/255

This PR adds an empty check for the `is_sticky` property to prevent the PHP errors. And also the PR fixes a few small phpcs errors in some comments.


includes/models/class-listing.php
```diff
@@ -792,7 +792,10 @@ public function update_plan( $plan = null, $args = array() ) {
 		}
 
 		$row['listing_id'] = $this->id;
-		$row['is_sticky']  = (int) $row['is_sticky'];
+
+		if ( ! empty( $row['is_sticky'] ) ) {
```
stephywells, line 796: Should this be isset instead? Just wondering what an empty value or false might do if it gets passed on.

sorinmarta: Great catch! From that line, I thought it was always either 0 or 1, but I checked in the `class-fee-plan.php` model and you are right, it can also be false so I pushed a commit to change that to isset.

stephywells: This still isn't quite right. The logic looks backwards now with !

sorinmarta: Right! 🤦‍♂️ 

Apologies for that.

stephywells (approved):

includes/models/class-listing.php
stephywells: Thanks! Looks good now!

PR #396, opened 2024-09-12
Check if the value is empty before adding https://

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/228

This PR adds a check for the value of the URL field, if it is empty it doesn't add the https:// protocol.


stephywells (approved):
Great catch! Thank you!

PR #397, opened 2024-09-23
Add no renewal email error message

sorinmarta:
closes https://github.com/Strategy11/business-directory-premium/issues/224

This PR adds a new error message if the user doesn't have any recurring email templates when they press the `Send renewal email` button.

### DIFF
--- a/includes/admin/class-admin.php
+++ b/includes/admin/class-admin.php
@@ -1075,17 +1075,7 @@ if ( ! class_exists( 'WPBDP_Admin' ) ) {
 						break;
 					}
 
-					$all_notices      = wpbdp_get_option( 'expiration-notices' );
-					$recurring_notice = false;
-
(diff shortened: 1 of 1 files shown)


includes/admin/class-admin.php
```diff
@@ -1075,6 +1075,21 @@ function handle_actions() {
 						break;
 					}
 
+					$all_notices      = wpbdp_get_option( 'expiration-notices' );
```
stephywells, line 1078: Can we move this new code into a separate function to avoid adding extra complexity?

sorinmarta: Good catch! Done, thanks!

stephywells (approved):
Thank you! This code looks great. Can we ask Garret to test it before merging?

garretlaxton (approved):
Sweet, this works great!

PR #398, opened 2024-09-24
Redirect inactive listings to main directory with notification

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/59

This PR implements a check for expired or inactive listings. When a user attempts to access such a listing, they are redirected to the main directory page, and a notification is displayed informing them that the listing is inactive.

### DIFF
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -96,33 +96,27 @@ class WPBDP__Dispatcher {
      * @return void
      */
     public function maybe_redirect_from_draft_listings() {
-        global $wp_query, $post, $wpdb;
+        global $wp_query;
 
-        if ( $post !== NULL ) {
+        if ( ! is_404() ) {
             return;
         }
 
-        $vars = $wp_query->query_vars;
+        $query_vars = $wp_query->query_vars;
 
-        if ( ! isset( $vars['post_type'] ) || $vars['post_type'] !== WPBDP_POST_TYPE ) {
+        if ( ! isset( $query_vars['post_type'] ) || $query_vars['post_type'] !== WPBDP_POST_TYPE ) {
             return;
         }
 
-        // We run a custom query here because we want to check for draft/pending listings and WP_Query doesn't return them.
-        $post = $wpdb->get_results(
-            $wpdb->prepare(
-                "
-                SELECT * FROM $wpdb->posts
-                WHERE post_type = %s
-                AND post_name = %s
-                AND post_status IN ('draft', 'pending')
-                ",
-                WPBDP_POST_TYPE,
-                $vars['name'],
+        $queried_posts = get_posts(
+            array(
+                'post_type'   => WPBDP_POST_TYPE,
+                'name'        => $query_vars['name'],
+                'post_status' => array( 'draft', 'pending' ),
             )
         );
 
-        if ( empty( $post ) ) {
+        if ( empty( $queried_posts ) ) {
             return;
         }
 
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -88,7 +88,7 @@ class WPBDP__Dispatcher {
 		return $template;
 	}
 
-	/**
+    /**
      * Redirect from draft listings to the main directory page.
      *
      * @since x.x.x
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -91,7 +91,7 @@ class WPBDP__Dispatcher {
     /**
      * Redirect from draft listings to the main directory page.
      *
-     * @since x.x.x
+     * @since x.x
      *
      * @return void
      */
--- a/includes/class-dispatcher.php
+++ b/includes/class-dispatcher.php
@@ -14,6 +14,7 @@ class WPBDP__Dispatcher {
 	public function __construct() {
 		add_action( 'wp', array( $this, '_lookup_current_view' ) );
 		add_action( 'template_redirect', array( $this, '_execute_view' ), 11 );
+		add_action( 'template_redirect', array( $this, 'maybe_redirect_from_draft_listings' ) );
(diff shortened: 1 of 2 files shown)


templates/main-box.tpl.php
```diff
@@ -7,6 +7,11 @@
 
 ?>
 <div id="wpbdp-main-box" class="wpbdp-main-box" data-breakpoints='{"tiny": [0,360], "small": [360,560], "medium": [560,710], "large": [710,999999]}' data-breakpoints-class-prefix="wpbdp-main-box">
+<?php
+    if ( isset( $_GET['inactive_listing'] ) && (int) $_GET['inactive_listing'] === 1 ) {
+	echo wpbdp_render_msg( _x( 'The listing you are trying to access is currently inactive or no longer available.', 'inactive-listing', 'business-directory-plugin' ), 'notice' );
```
stephywells, line 12: Can we remove the 'inactive-listing' part here? I don't think it's helpful in this case so it would be nice to keep the code more simple.

includes/class-dispatcher.php
```diff
@@ -87,6 +88,48 @@ public function _execute_view( $template ) {
...
+     * @since x.x.x
+     *
+     * @return void
+     */
+    public function maybe_redirect_from_draft_listings() {
+        global $wp_query, $post, $wpdb;
+
+        if ( $post !== NULL ) {
+            return;
+        }
```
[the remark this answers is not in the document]
stephywells: @sorinmarta Would ! $post do the job here?

includes/class-dispatcher.php
```diff
@@ -87,6 +88,48 @@ public function _execute_view( $template ) {
...
+     * Redirect from draft listings to the main directory page.
+     *
+     * @since x.x.x
+     *
+     * @return void
+     */
+    public function maybe_redirect_from_draft_listings() {
+        global $wp_query, $post, $wpdb;
+
+        if ( $post !== NULL ) {
```
stephywells, line 101: Could we return immediately if we only run this if is_404()? I'm not sure if it would return true at this point or not, but I think it might.

includes/class-dispatcher.php
```diff
@@ -87,6 +88,48 @@ public function _execute_view( $template ) {
...
+        }
+
+        $vars = $wp_query->query_vars;
+
+        if ( ! isset( $vars['post_type'] ) || $vars['post_type'] !== WPBDP_POST_TYPE ) {
+            return;
+        }
+
+        // We run a custom query here because we want to check for draft/pending listings and WP_Query doesn't return them.
+        $post = $wpdb->get_results(
```
stephywells, line 112: It looks like we might have the rules not mentioning overriding globals. Setting $post here will cause an issue later when we ratchet up the phpcs checks. Can we use a different var name here?

includes/class-dispatcher.php
```diff
@@ -87,6 +88,48 @@ public function _execute_view( $template ) {
...
+            $wpdb->prepare(
+                "
+                SELECT * FROM $wpdb->posts
```
stephywells, line 115: Is there a reason not to use get_posts()?

stephywells (changes requested):
I haven't tested this, but it looks great. Just a few small questions.

templates/main-box.tpl.php
sorinmarta: Do you mean the context property of the _x function? That will result in a phpcs error because it is a [required parameter](https://developer.wordpress.org/reference/functions/_x/) of the function.

Or do you mean the query parameter as a whole? I guess we could change the approach to a transient, but I opted for the query parameter to avoid the cache/database calls.

includes/class-dispatcher.php
sorinmarta: Oh 🤦‍♂️ , I didn't realize that I was overwriting the global `$post`, I was too focused on why `get_posts` was not retrieving the post. But I'll answer that in the next comment.

includes/class-dispatcher.php
sorinmarta: I'll address the question above here because I can't comment on that one.

>Would ! $post do the job here?

It would but without the negation 😄 . I'll change it to that, thanks!

>Could we return immediately if we only run this if is_404()? I'm not sure if it would return true at this point or not, but I think it might.

Good idea! That would remove the need for the $post global!

includes/class-dispatcher.php
sorinmarta: I was definitely doing something wrong and not realizing what because when I worked on it first no matter what I tried with `get_posts` it was always returning an empty array so eventually I went for a custom query.

But now when I tried it again after a few hours break it worked so it was just me being too tired and making silly mistakes. 

I changed that to using `get_posts`. Thanks @stephywells!

templates/main-box.tpl.php
stephywells: Sorry. I meant switch _x() to __() since the extra description isn't helpful to a translator. I frequently make this switch in BD because it ends up creating a bunch of extra strings that need translation when they are used in the same way.

includes/class-dispatcher.php
```diff
@@ -87,6 +88,42 @@ public function _execute_view( $template ) {
 		return $template;
 	}
 
+    /**
+     * Redirect from draft listings to the main directory page.
+     *
+     * @since x.x.x
```
stephywells, line 94: This should be x.x so it gets replaced correctly. Here's where the replacement is handled:
https://github.com/Strategy11/business-directory-plugin/blob/master/Gruntfile.js#L210

stephywells (commented):
Thanks for all the updates! This looks much cleaner and easier to follow now.

sorinmarta:
Done! Please let me know if you have any other suggestions @stephywells! Thanks!

stephywells (approved):
This looks perfect! Awesome!

PR #399, opened 2024-09-27
Update CodeCeption composer packages

sorinmarta:
This PR updates the composer packages to fix CodeCeption issues.


sorinmarta:
Fixed in https://github.com/Strategy11/business-directory-regions/pull/56

PR #400, opened 2024-09-27
Fix priority conflict for the_content hooks

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/225

This PR fixes the priority conflict for `the_content` hooks.

### DIFF
--- a/business-directory-plugin.php
+++ b/business-directory-plugin.php
@@ -121,10 +121,17 @@ function wpbdp() {
 	return $instance;
 }
 
+// Increase the priority value for AddToAny widgets.
+add_filter(
+    'addtoany_content_priority', function () {
+    return 1000;
+}, 100);
+
+// Increase the priority value for Social Share Buttons widgets.
 add_filter(
     'ssb_the_content_priority', function () {
         return 1000;
-}, 1);
+}, 100);
 
 
 // For backwards compatibility.
--- a/business-directory-plugin.php
+++ b/business-directory-plugin.php
@@ -121,12 +121,6 @@ function wpbdp() {
 	return $instance;
 }
 
-// Increase the priority value for AddToAny widgets.
-add_filter(
-    'addtoany_content_priority', function () {
-    return 1000;
-}, 100);
-
 // Increase the priority value for Social Share Buttons widgets.
 add_filter(
     'ssb_the_content_priority', function () {
--- a/business-directory-plugin.php
+++ b/business-directory-plugin.php
@@ -121,6 +121,11 @@ function wpbdp() {
 	return $instance;
 }
 
+add_filter(
+    'ssb_the_content_priority', function () {
+        return 1000;
+}, 1);
+
 
 // For backwards compatibility.
 $GLOBALS['wpbdp'] = wpbdp();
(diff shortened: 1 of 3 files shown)


business-directory-plugin.php
```diff
@@ -121,6 +121,18 @@ function wpbdp() {
 	return $instance;
 }
 
+// Increase the priority value for AddToAny widgets.
```
stephywells, line 124: Can we move this to a different file to avoid bulking this one up? This looks like a good spot for it:
https://github.com/Strategy11/business-directory-plugin/blob/master/includes/compatibility/class-compat.php

stephywells (changes requested):
Great find!!! Yay!! This looks like a great solution.

business-directory-plugin.php
sorinmarta: Done, I moved it into a function there. But the one for SSB has to stay in the main plugin file, otherwise, the filter will be executed before we can increase its value.

stephywells (changes requested):
Thank you!

I'm also seeing this line:
remove_filter( 'the_content', array( $this, 'display_view_in_content' ), 999 );

It looks like this needs adjustment too?

sorinmarta:
>I'm also seeing this line:
remove_filter( 'the_content', array( $this, 'display_view_in_content' ), 999 );
It looks like this needs adjustment too?

You are right, it needed adjustment, and it also proved that the conflict was not with the same priority value. Banago was partially right about the `remove_filter` line.

Changing the priorities to different values prevented the issue because when I edited it I added by mistake different values for the add_filter and remove_filter. So basically the filter was never removed.

For some reason that I can't understand, [this line](https://github.com/Strategy11/business-directory-plugin/blob/master/includes/class-wordpress-template-integration.php#L253) removes the filter for non-tax and it also prevents the execution of the filter with the next priority. The filter still shows up in the filter stack, but it doesn't execute. Which is why I think Banago thought we have to remove it.

So initially I thought about refactoring the way we render the content into 2 separate methods. One for taxonomies and one for the regular pages. But I couldn't even reproduce the Astra issues that required those lines in the first place. Most likely I don't have the right settings in place to cause the error.

In the end, I decided to keep it simple and add another blank filter after the one we intend to remove right before we remove it, that way that filter will be prevented and all the others following after that will be executed correctly.

includes/class-wordpress-template-integration.php
```diff
@@ -250,6 +250,11 @@ public function display_view_in_content( $content = '' ) {
 		$is_tax = is_tax();
 
 		if ( ! $is_tax ) {
+            // This filter is added because of an issue with remove_filter. 
+            // See: https://github.com/Strategy11/business-directory-plugin/pull/400
+            add_filter( 'the_content', fn( $content ) => $content, 1000 );
```
stephywells, line 255: Looks like this format requires php 7.4, but we are still supporting previous versions.

stephywells (changes requested):
Great job finding a workaround! Weird, but this looks like a great way to minimize conflicts or regressions.

includes/class-wordpress-template-integration.php
sorinmarta: Right, I thought we support 7.4, but we support 7.0. I'll keep that in mind. Thanks @stephywells!

stephywells (approved):
Thanks for the fix!

PR #401, opened 2024-10-02
This PR adds echo to the flag data for listing reports

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/256

This PR adds `echo` before the listing details in the admin dashboard.

### DIFF
--- a/templates/admin/metaboxes-listing-flagging-row.tpl.php
+++ b/templates/admin/metaboxes-listing-flagging-row.tpl.php
@@ -39,14 +39,14 @@ endif;
 		</div>
 		<div class="report-reasons">
 			<?php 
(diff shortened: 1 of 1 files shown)


stephywells (changes requested):
Oops!! Thanks for stepping in and making the quick fix.

It looks like some of these changed lines are using spaces now instead of tabs. Once we update that we can merge it.

sorinmarta:
Good catch! I wonder why they had spaces, my editor is configured to use tabs. And also I wonder why phpcs didn't sniff that.

I'm still a newbie at using VSCode so most likely I messed up something 🤦‍♂️

garretlaxton (approved):
Thanks for looking at this so quickly! This fix works great!

PR #402, opened 2024-10-04
New/stripe

Crabcyborg:


Crabcyborg:
@garretlaxton This update should now be ready enough for QA.

It isn't 100% complete, but it's really close. I put a "Current known issues" section in about a toggle issue I haven't figured out yet.

To start, go to Payment Settings, go to Stripe, and connect using the buttons like you would with Formidable. The process should be really similar, though there's a "Enable Stripe?" toggle here in BD that is causing some issues.

I have a couple other TODO items I listed as well.

I'm still learning the BD plugin, so I'm eager to have someone with more experience testing this to make sure I didn't miss anything.

Thank you!

garretlaxton:
I don't see the upgrade banner as seen in your image. I figured I would see this when the Premium and Stripe add-ons are installed?
image


I also saw this warning in my logs after making a payment.
`PHP Warning:  Undefined property: stdClass::$postal_code in /public_html/bd/wp-content/plugins/business-directory-plugin/includes/gateways/stripe/controllers/classStrpEventsController.php on line 334`

Crabcyborg:
> PHP Warning:  Undefined property: stdClass::$postal_code in /public_html/bd/wp-content/plugins/business-directory-plugin/includes/gateways/stripe/controllers/classStrpEventsController.php on line 334

Thanks for catching that @garretlaxton! I made a mistake in the API code and had `zip` instead of `postal_code`. This should be fixed now.

> PHP Warning:  Undefined property: stdClass::$payment_intent in /public_html/bd/wp-content/plugins/business-directory-plugin/includes/gateways/stripe/controllers/classStrpEventsController.php on line 314

Thanks! This warning should stop happening now as well with the latest commits.

> I don't see the upgrade banner as seen in your image. I figured I would see this when the Premium and Stripe add-ons are installed?

This is behind a couple conditions. It goes behind a `license_includes_stripe_fees` function, so it's only shown when there are fees. If you have access to the Stripe add-on, you shouldn't see it. The check at the moment really just checks the Add-ons API and returns true for a license that does have access to the Stripe add-on.

Crabcyborg:
@garretlaxton I've also pushed a few new changes.

- Now when the BD Stripe module is also active, it will be automatically unloaded if any of the following is true.
    - Both Stripe Connect test and live modes are set up.
    - No API keys are set yet.
- BD Stripe Lite will not be loaded instead if BD Stripe module is active, and API keys are defined.

garretlaxton:
The payment history is showing the wrong dates when I renew an expired listing. This is a non-recurring plan, and the Stripe add-on was disabled when I did the renewal payment.

It shows that the listing was renewed on December 4th at 1am UTC in the listing timeline section when editing a listing, which is correct.
image

But the payment history and checkout page show November 19th for some reason.
image
image

garretlaxton:
Your latest changes with the add-on functionality work perfectly. Should we move the `Checkout Window Title` option to the Stripe add-on?

Crabcyborg:
@garretlaxton I've made some more changes.

I haven't figured out the issue with the wrong date appearing yet. Does this happen to be an issue with the Stripe add-on as well? The code isn't very different. I'm waiting for a list to really expire today so I can check tomorrow.

What I have done
- I added Authorize.Net back in. This will only appear as an option if you already had it configured in the previous releases. Or you can show it with `add_filter( 'wpbdp_include_authorize_net_gateway', '__return_true' );`.
- I noticed another PHP warning for another missing `customer` property on the invoice object.
- The list of Payment Gateways at the bottom of the list of Plans now shows "Set Up" for Stripe instead of "Upgrade".
- The Connected status will now display a checkmark when it is active.

Crabcyborg:
> Should we move the Checkout Window Title option to the Stripe add-on?

@garretlaxton Regarding this question, the setting should work the same as it does in the Stripe module. The setting only works for one time payments, but it should appear on the checkout session page when using Stripe Lite as well.

image

includes/gateways/stripe/helpers/classStrpAppHelper.php
```diff
@@ -0,0 +1,65 @@
...
+	 * @return string
+	 *
+	 * @psalm-return 'live'|'test'
+	 */
+	public static function active_mode() {
+		return wpbdp_get_option( 'payments-test-mode' ) ? 'test' : 'live';
+	}
+
+	/**
+	 * Add education about Stripe fees.
```
sorinmarta, line 23: We should be consistent with doc comments for all new methods.

includes/gateways/stripe/js/connect-settings.js
```diff
@@ -0,0 +1,132 @@
+( function() {
+	function setupStripeConnectListener() {
+		onclickPreventDefault( '#wpbdp_disconnect_stripe', handleStripeDisconnectClick );
+		onclickPreventDefault( '#wpbdp_reauth_stripe', handleStripeReauthClick );
+		onclickPreventDefault( '#wpbdp_connect_with_oauth', handleConnectWithOauth );
+		renderStripeConnectSettingsButton();
+	}
+
+	function renderStripeConnectSettingsButton() {
+		var container = document.getElementById( 'wpbdp_strp_settings_container' );
```
sorinmarta, line 10: I think we are safe with using either `const` or `let` for all variables

includes/gateways/stripe/js/connect-settings.js
```diff
@@ -0,0 +1,132 @@
...
+					messageEle.innerHTML = data.message;
+					messageEle.classList.remove( 'wpbdp-hidden' );
+				}
+				renderStripeConnectSettingsButton();
+			}
+		);
+	}
+
+	function strpSettingsAjaxRequest( action, success, testMode, fail ) {
+		var data = {
```
sorinmarta, line 85: `const` would be safe here too.

includes/gateways/stripe/js/connect-settings.js
```diff
@@ -0,0 +1,132 @@
...
+			action: action,
+			testMode: testMode,
+			nonce: wpbdp_global.nonce
+		};
+		postAjax( data, success, fail );
+	}
+
+	function postAjax( data, success, fail ) {
+		var xmlHttp, params;
```
sorinmarta, line 94: `const` would be safe here too.

We could do something like this:
```js
const xmlHttp = new XMLHttpRequest();
const params = typeof data === 'string' ? data : Object.keys( data ).map(
			function( k ) {
				return encodeURIComponent( k ) + '=' + encodeURIComponent( data[ k ]);
			}
		).join( '&' );
```

Although, perhaps we should use jQuery.post() to make the call for consistency and readability.

sorinmarta (commented):
Thanks @Crabcyborg! Amazing work! 🚀 

I left a few comments, but only tiny details.

includes/gateways/stripe/js/connect-settings.js
Crabcyborg: Thanks Sorin!

This is fairly old code, copied from Formidable Stripe.

The main reason it's does with XHR is to avoid jQuery. At least in Formidable, our long term goal is to remove our jQuery dependency.

If we rewrite it, I would opt to use `fetch` instead. I just don't know if it's worth rewriting since it already works.

includes/class-payment-gateways.php
```diff
@@ -37,6 +48,86 @@ public function load_gateways() {
...
+	 * @return bool
+	 */
+	private function settings_key_exists( $keys ) {
+		$settings = get_option( 'wpbdp_settings' );
+		if ( ! is_array( $settings ) ) {
+			return false;
+		}
+
+		foreach ( $keys as $key ) {
+			if ( ! empty( $settings[ $key ] ) ) {
```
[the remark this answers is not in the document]
Crabcyborg: We really only want to know if one key exists for the purpose of this check.

It's really just to determine if we should hide their settings or not. If something/anything is set, we don't want to hide it.

assets/js/checkout.js
```diff
@@ -47,7 +47,44 @@ jQuery(function($) {
...
+            stripe.redirectToCheckout({
+                // Make the id field from the Checkout Session creation API response
+                // available to this file, so you can provide it as parameter here
+                // instead of the {{CHECKOUT_SESSION_ID}} placeholder.
+                sessionId: configuration.sessionId
+            }).then(function ( result ) {
+                // If `redirectToCheckout` fails due to a browser or network
+                // error, display the localized error message to your customer
+                // using `result.error.message`.
+            });
```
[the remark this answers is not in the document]
Crabcyborg: Thanks @sorinmarta. I agree. Do you know how we normally show errors in BD? I just don't really know how this would normally be done as I'm new to the plugin.

includes/gateways/stripe/js/connect-settings.js
Crabcyborg: I updated these to const for now.

includes/gateways/stripe/helpers/classStrpAppHelper.php
Crabcyborg: Thanks! I've gone through and added comments to every new function with more consistency.

includes/gateways/stripe/controllers/classStrpEventsController.php
```diff
@@ -0,0 +1,377 @@
...
+			if ( $this->should_skip_event( $event_id ) ) {
+				continue;
+			}
+
+			set_transient( 'wpbdp_last_process_' . $event_id, time(), 60 );
+
+			$this->event = WPBDPStrpConnectHelper::get_event( $event_id );
+			if ( is_object( $this->event ) ) {
+				$this->handle_event();
+				$this->track_handled_event( $event_id );
```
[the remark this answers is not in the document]
Crabcyborg: I think I prefer it this way. It reduces the complexity of each of the functions having the logic separated and allows us to test each thing in isolation (ie. I could write tests for `handle_event` and `track_handled_event` separately instead of having to test both things together).

Since `handle_event` is a larger function I think it would also be easier to miss that `track_handled_event` is called at the bottom of the `switch/case`.

includes/gateways/stripe/controllers/classStrpEventsController.php
```diff
@@ -0,0 +1,377 @@
...
+		if ( $failed_count >= $maximum_retries ) {
+			$this->track_handled_event( $event_id );
+		} else {
+			set_transient( $transient_name, $failed_count, 4 * DAY_IN_SECONDS );
+		}
+	}
+
+	/**
+	 * Track an event to no longer process.
+	 * This is called for successful events, and also for failed events after a number of retries.
```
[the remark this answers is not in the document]
Crabcyborg: I don't think I would ever remove comments for consistency.

I tried to go through and add some more comments to most of the functions.

includes/gateways/stripe/js/connect-settings.js
Crabcyborg: I updated this again to use `fetch` now. It was pretty simple with AI https://github.com/Strategy11/business-directory-plugin/pull/402/commits/7eb7300aa426b69c2474fb7118cd055687a20c86

Crabcyborg: And another commit to simplify it more https://github.com/Strategy11/business-directory-plugin/pull/402/commits/289b9dbb203e9c466a91895d4acd6c303587e5de

includes/class-modules.php
```diff
@@ -35,6 +39,68 @@ private function register_modules() {
...
+		}
+
+		$keys = array(
+			'stripe-test-publishable-key',
+			'stripe-test-secret-key',
+			'stripe-live-publishable-key',
+			'stripe-live-secret-key',
+		);
+		foreach ( $keys as $key ) {
+			if ( ! empty( $settings[ $key ] ) ) {
```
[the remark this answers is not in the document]
Crabcyborg: We don't need to know if all of them are set. I really just wants to know if something is filled in.

assets/js/checkout.js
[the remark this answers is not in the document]
Crabcyborg: I updated this for now to use `console.error` like you mention.

includes/class-payment-gateways.php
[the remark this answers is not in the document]
sorinmarta: Understood!

I think we could add it as a public method in [this settings class](https://github.com/Strategy11/business-directory-plugin/blob/master/includes/admin/settings/class-settings.php). It would make sense and it would also be easy to use through the `$wpbdp` global.

Something like this:
```php
$wpbdp->settings->any_setting_exists( $keys );
```

The current functions are used twice in `class-payment-gateways.php` and the other one is used only once in `class-modules.php`.

sorinmarta:
Thanks @Crabcyborg! Besides the settings key comment, I think everything else is ready to go! 🚀 

But even that can be done in a future update to speed up the release as it doesn't influence any functionality.

Amazing work! 🚀

garretlaxton:
This is looking great! I wasn't able to replicate the previous issue anymore. I want to let a few listings expire and renew tomorrow morning, so I'll report back then.

includes/class-payment-gateways.php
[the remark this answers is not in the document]
Crabcyborg: Thank you @sorinmarta! That's a great improvement.

Updated in https://github.com/Strategy11/business-directory-plugin/pull/402/commits/f16ad78bf9ac2980e1c8b38f4f8ed6b9bb400518

And also https://github.com/Strategy11/business-directory-plugin/pull/402/commits/088aecd2dd19965d2ec677e7a87772b1181d04a6

Crabcyborg:
@sorinmarta I think I'm complete now with making changes. The remaining items on my list are all testing related.

I added some validation so it will show a popup in the corner if you have Stripe enabled but have not connected live payments yet.

I also added a check to make sure the lite uuid is never the easy to guess hash, forcing the new unique ID format. This could mean that tracking would start to track someone as a new user, but I don't believe out tracking in BD is functional at the moment anyway.

Thank you for your help reviewing!

My next question - When do you think we could release this? I'll want you to do the release as you normally would.

We have a deadline with Stripe. It sounds like the API key functionality will possibly even cease to work by the deadline. We need it live by at least December 31st.

I think we could possibly aim for Monday, the 23rd. With the holidays, I would probably either aim for then or on the following Monday, the 30th, which would be cutting it kind of close.

What do you think?

**_One more thing_** - do you think we should add any extra messaging in the plugin to draw attention to the new Stripe integration? Something that announces like "New! - Stripe Connect is now available in the BD Lite! Make money from listings for free with a 3% fee. Click here to learn more".

sorinmarta (approved):
@Crabcyborg Monday the 23rd sounds good to me. Garret's idea of adding a corner notification sounds good.

You can see how we've added that in [this class](https://github.com/Strategy11/business-directory-plugin/blob/master/includes/licensing.php).

And perhaps we can add the new notification code [here](https://github.com/Strategy11/business-directory-plugin/blob/master/includes/class-payment-gateways.php#L111).

Crabcyborg:
Thank you @sorinmarta!

I added a new commit with some new messaging https://github.com/Strategy11/business-directory-plugin/pull/402/commits/90f41c889a60b836d8cb69c36b2f489d4dc528f9

I've added 3 different messages based on various conditions.

Screen Shot 2024-12-20 at 10 39 26 AM
Screen Shot 2024-12-20 at 10 39 19 AM
Screen Shot 2024-12-20 at 10 38 53 AM

Crabcyborg:
Thank you @garretlaxton!

🚀

PR #405, opened 2024-10-10
Prevent caching null

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/257

This PR adds a check before the `set_cache` method to make sure we don't save an empty value in cache if that is enabled when the `check_cache` is called.

And also it adds the `$allow_empty` parameter to `wpbdp_get_post_by_id_or_slug()` to give the option to ignore the empty values.

### DIFF
--- a/includes/utils.php
+++ b/includes/utils.php
@@ -107,7 +107,7 @@ class WPBDP_Utils {
 	 *
 	 * @return mixed $results The cache or query results
 	 */
-	public static function check_cache( $args ) {
+	public static function check_cache( $args, $allow_empty = true ) {
 		$defaults = array(
 			'cache_key' => '',
 			'group'     => '',
@@ -144,7 +144,7 @@ class WPBDP_Utils {
 			}
 		}
 
-		if ( ! empty( $results ) ) {
+		if ( $allow_empty || ! empty( $results ) ) {
 			self::set_cache( $args['cache_key'], $results, $args['group'], $args['time'] );
 		}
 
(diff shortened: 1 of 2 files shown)


includes/utils.php
```diff
@@ -144,7 +144,9 @@ public static function check_cache( $args ) {
 			}
 		}
 
-		self::set_cache( $args['cache_key'], $results, $args['group'], $args['time'] );
+		if ( ! empty( $results ) ) {
```
shervElmi: Hi @sorinmarta,

Thanks for your work on this.

We should move the validation inside the `self::set_cache` method as it's more future-proof and will offer safer usage elsewhere in the future.

shervElmi (approved):
@sorinmarta, it looks great to me. however, I didn’t do QA tests. I’ve also added Garret for this purpose.

🚀

includes/utils.php
sorinmarta: Thanks for the review @shervElmi! I didn't expect you to be so fast. 😄

While you were reviewing I was thinking about it and realized that checking for empty for every kind of usage may cause a performance impact for things that should be stored as empty.

Like an empty category for example. It will always run the query and could definitely be optimized with cache.

So I decided to switch to a parameter that can then be used only where empty values should not be stored.

Good point with moving that to `set_cache` as that's also a public method. I'll change it to that.

Thanks!

sorinmarta:
Thanks @shervElmi!

@garretlaxton I'll also make a PR for business-directory-premium that will make use of this change to fix the Elementor issue.

That will need to be tested as this PR doesn't really change anything in the UX. I'll make sure to add instructions on how to test there.

PR #410, opened 2024-10-23
Fix deprecated use of get_class in PHP 8.4

Crabcyborg:
I'm running PHP 8.4 RC2 right now locally.

This PR fixes a new deprecated message that is introduced in PHP 8.4.

> PHP Deprecated:  Calling get_class() without arguments is deprecated in /var/www/src/wp-content/plugins/business-directory-plugin/includes/models/class-payment.php on line 422

Calling `get_class()` with no arguments in a static function appears to work the same as using `__CLASS__`.


sorinmarta (approved):
Looks good to me, thanks @Crabcyborg!

PR #411, opened 2024-10-23
Fix deprecated null param in number format in PHP 8.4

Crabcyborg:
Using PHP 8.4 RC2, I see this deprecated message.

This update checks if `$num` is null, and if it is, converts it to a safe float value first.

> PHP Deprecated:  number_format(): Passing null to parameter #1 ($num) of type float is deprecated in /var/www/src/wp-content/plugins/business-directory-plugin/includes/helpers/class-currency-helper.php on line 62


sorinmarta (approved):
Thanks @Crabcyborg!

PR #412, opened 2024-10-25
Fix php warning on email templates page

Crabcyborg:
When loading the email template settings page (/wp-admin/admin.php?page=wpbdp_settings&tab=email&subtab=email_templates), I see this PHP warning.

> PHP Warning:  Trying to access array offset on false in /var/www/src/wp-content/plugins/business-directory-plugin/includes/admin/controllers/class-settings-admin.php on line 616

This update just checks that `$setting['default']` is actually an array first, and that the key set. In my case, it was `false`.

There might be a better way to do this, or we might not want to use an empty string here as a fallback? I'm just avoiding the warning and haven't looked too much into this.

### DIFF
--- a/includes/admin/controllers/class-settings-admin.php
+++ b/includes/admin/controllers/class-settings-admin.php
@@ -610,10 +610,15 @@ class WPBDP__Settings_Admin {
 		$this->setting_text_callback( $setting, $value );
 	}
 
+	/**
+	 * @param array|false $setting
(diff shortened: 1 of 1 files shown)


sorinmarta (approved):
Thanks @Crabcyborg! I think that's a good workaround to prevent the error and the empty string makes sense to me.

Crabcyborg:
Thanks Sorin!

PR #416, opened 2024-11-26
Add black friday menu item

sorinmarta:
This PR adds a promotional link for Black Friday to the admin sub menu


includes/admin/class-admin.php
```diff
@@ -404,6 +404,92 @@ function admin_menu() {
...
+		}
+
+		/**
+		 * Redirect to Black Friday sales page when the menu item is clicked.
+		 *
+		 * @since x.x
+		 *
+		 * @return void
+		 */
+		public function redirect_blackfriday() {
```
Crabcyborg, line 487: @sorinmarta It looks like you don't need this function, since you went with directly adding the URL to the submenu.

assets/css/less/admin.less
```diff
@@ -57,6 +57,10 @@
     }
 }
 
+#adminmenu .wpbdp-orange-text {
+	color: #F47449;
```
shervElmi, line 61: Is there a specific reason why you didn't add this color to `colors.less`?

includes/admin/class-admin.php
```diff
@@ -404,6 +404,79 @@ function admin_menu() {
...
+			global $submenu;
+			$submenu[ $this->menu_id ][] = array(
+				$black_friday_menu_label,
+				'manage_options',
+				wpbdp_admin_upgrade_link( 
+					array( 
+						'medium'  => 'black-friday-submenu', 
+						'content' => 'black-friday-submenu',
+					), 
+					'black-friday' ),
```
shervElmi, line 436: Could you move the closing parenthesis ")" to a new line?

shervElmi (approved):
It looks good to me, @sorinmarta. Thanks! 🚀

I left two comments that are minor.

assets/css/less/admin.less
sorinmarta: That file contains only variables. And the color of that link is temporary just for this Black Friday so I didn't really see a reason for it to be a variable.

PR #424, opened 2025-01-14
Prevent Elementor from loading element cache on shortcodes

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/269

This PR checks if Elementor is active and if the current post has one of our shortcodes. If it has any shortcode it will delete the current element cache value to make sure we always render the right page.

### DIFF
--- a/includes/compatibility/class-compat.php
+++ b/includes/compatibility/class-compat.php
@@ -149,14 +149,10 @@ class WPBDP_Compat {
 		}
 
 		$cache_key  = Elementor\Core\Base\Document::CACHE_META_KEY;
-		$shortcodes = array_keys( $wpbdp->shortcodes->get_shortcodes() );
+		$shortcodes = get_shortcode_regex( array_keys( $wpbdp->shortcodes->get_shortcodes() ) );
 
-		foreach ( $shortcodes as $shortcode ) {
-			if ( has_shortcode( $post->post_content, $shortcode ) ) {
-				delete_post_meta( $post->ID, $cache_key );
-			}
-
-			break;
+		if ( preg_match( "/$shortcodes/", $post->post_content ) ) {
+		    delete_post_meta( $post->ID, $cache_key );
 		}
 	}
 
--- a/includes/compatibility/class-compat.php
+++ b/includes/compatibility/class-compat.php
@@ -64,6 +64,11 @@ class WPBDP_Compat {
 			add_filter(
                 'wpf_skip_auto_login', array( $this, 'wp_fusion_skip_auto_login' ), 20 );
 		}
+
+		// Delete Elementor element caching for shortcodes.
+		if ( class_exists( 'Elementor\Core\Base\Document' ) ) {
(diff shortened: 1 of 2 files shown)


includes/compatibility/class-compat.php
```diff
@@ -129,6 +134,32 @@
...
+	 * @return void
+	 */
+	public function prevent_shortcodes_elementor_element_cache() {
+		global $post, $wpbdp;
+
+		if ( ! $post ) {
+			return;
+		}
+
+		$cache_key = Elementor\Core\Base\Document::CACHE_META_KEY;
```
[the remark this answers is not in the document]
sorinmarta: We check for that before the WordPress action is registered

includes/compatibility/class-compat.php
```diff
@@ -129,6 +134,28 @@ public function wp_fusion_skip_auto_login( $skip_auto_login ) {
...
+	 */
+	public function prevent_shortcodes_elementor_element_cache() {
+		global $post, $wpbdp;
+
+		if ( ! $post ) {
+			return;
+		}
+
+		$cache_key  = Elementor\Core\Base\Document::CACHE_META_KEY;
+		$shortcodes = get_shortcode_regex( array_keys( $wpbdp->shortcodes->get_shortcodes() ) );
```
shervElmi, line 152: We can stop the process earlier if there are no shortcodes, which will optimize this method further. Currently, `post_content` is always being processed inside `preg_match`, and for large data, this can lead to performance issues as the entire content is passed to the regex. Restructuring the method, as shown below, can help improve efficiency:

```php
	/**
	 * Checks if the current post content has a BD shortcode and deletes the Elementor cache.
	 *
	 * @since x.x
	 *
	 * @return void
	 */
	public function prevent_shortcodes_elementor_element_cache() {
		global $post, $wpbdp;

		if ( ! $post ) {
			return;
		}

		$shortcodes = $wpbdp->shortcodes->get_shortcodes();

		if ( empty( $shortcodes ) ) {
			return;
		}

		$cache_key       = Elementor\Core\Base\Document::CACHE_META_KEY;
		$shortcode_regex = get_shortcode_regex( array_keys( $shortcodes ) );

		if ( preg_match( "/$shortcode_regex/", $post->post_content ) ) {
			delete_post_meta( $post->ID, $cache_key );
		}
	}
```

shervElmi (approved):
Thanks @sorinmarta, it looks great to me! 🚀 There is just one point I’d like to highlight below.

includes/compatibility/class-compat.php
sorinmarta: `$shortcodes` will never be empty, that's a list of all available shortcode names from our plugin and all of them are registered before `template_redirect`

Do you think there will be scenarios where the shortcodes won't be registered in time for that hook?

sorinmarta: Thanks @shervElmi!

PR #426, opened 2025-01-15
Remove spaces from sorting

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/263

This PR adds a filter to the query order by clause to remove spaces from the value to improve sorting both in the initial query that renders the first page and if the user changes the sorting order.

### DIFF
--- a/includes/class-query-integration.php
+++ b/includes/class-query-integration.php
@@ -221,8 +221,8 @@ class WPBDP__Query_Integration {
 		}
 
 		$field = explode( ' ', $orderby )[0];
-		
-		return "REPLACE({$field}, ' ', '') " . $query->get( 'order', 'ASC' );
+
+		return $this->get_space_replace( $field, $query->get( 'order', 'ASC' ) );
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


includes/class-query-integration.php
```diff
@@ -200,6 +200,29 @@ public function pre_get_posts( &$query ) {
...
+	 * 
+	 * @since x.x
+	 *
+	 * @param string   $orderby - The orderby clause.
+	 * @param WP_Query $query - The current query object.
+	 *
+	 * @return string
+	 */
+	public function remove_spaces_from_order_by( $orderby, $query ) {
+		if ( ! isset( $query->wpbdp_our_query ) || ! $query->wpbdp_our_query ) {
```
shervElmi, line 219: Could you use the `empty` function here, as it includes both `isset` and value checks?

```php
if ( empty( $query->wpbdp_our_query ) )
```

includes/class-query-integration.php
```diff
@@ -466,12 +489,26 @@ public function sortbar_orderby( $orderby ) {
...
+	 * 
+	 * @since x.x
+	 *
+	 * @param string $field - The field to remove spaces from.
+	 * @param string $order - The order to apply.
+	 * 
+	 * @return string
+	 */
+	private function get_space_replace( $field, $order ) {
+		return "REPLACE({$field}, ' ', '') " . ' ' . $order;
```
shervElmi, line 509: Could you add spaces around the `REPLACE` function parentheses?

shervElmi (approved):
Thanks @sorinmarta, It looks great to me! 🚀

PR #427, opened 2025-01-17
Store the field label to allow for translations

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/260

This PR changes the DOM rendering for the multi-select and radio field types to use the label value in the value attribute.

This way we can store and render the translated values on the listing page.

It also saves the selected value of those fields as separate post meta so that we know which one is selected/checked no matter what translated value is stored.

### DIFF
--- a/includes/abstracts/class-form-field-type.php
+++ b/includes/abstracts/class-form-field-type.php
@@ -116,7 +116,7 @@ class WPBDP_Form_Field_Type {
 	 *
 	 * @param WPBDP_Form_Field_Type $field The field object.
 	 * 
-	 * @return null|string
+	 * @return string|null
 	 */
 	public function get_field_selected_value( &$field ) {
 		// phpcs:ignore
--- a/includes/abstracts/class-form-field-type.php
+++ b/includes/abstracts/class-form-field-type.php
@@ -109,6 +109,33 @@ class WPBDP_Form_Field_Type {
 		return $value;
 	}
 
+	/**
+	 * Retrieve the option name for the selected value.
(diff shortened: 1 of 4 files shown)


includes/abstracts/class-form-field-type.php
```diff
@@ -109,6 +109,33 @@ public function get_field_value( &$field, $post_id ) {
...
+		$post_param    = wpbdp_get_var( array( 'param' => 'post' ), 'get' );
+
+		if ( empty( $listing_param ) && empty( $post_param ) ) {
+			return null;
+		}
+
+		$post_id = ! empty( $listing_param ) ? intval( $listing_param ) : intval( $post_param );
+		$value   = get_post_meta( $post_id, '_wpbdp[fields][' . $field->get_id() . ']_selected', true );
+
+		if ( false === $value ) {
```
shervElmi, line 132: We can also write the return like:

```php
return false !== $value ? $value : null;
```

includes/fields/class-fieldtypes-radiobutton.php
```diff
@@ -79,6 +86,12 @@ public function render_field_inner( &$field, $value, $context, &$extra = null, $
 		return $html;
 	}
 
+	public function store_field_value( &$field, $post_id, $value ) {
```
shervElmi, line 89: Could you add a PHPDoc comment here as well?

shervElmi (approved):
Thanks @sorinmarta, it looks great to me! 🚀

PR #429, opened 2025-01-23
Add check for field options before storing their value

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/284

This PR makes sure the multi-options fields have options before storing their value.


shervElmi (approved):
Thanks @sorinmarta, it looks good to me! 🚀

PR #431, opened 2025-01-27
Change the new multi value select to run only for meta fields

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/286

This PR changes the new way we added to store the value of multi value fields to only run for meta fields.

### DIFF
--- a/includes/fields/class-fieldtypes-radiobutton.php
+++ b/includes/fields/class-fieldtypes-radiobutton.php
@@ -72,26 +72,15 @@ class WPBDP_FieldTypes_RadioButton extends WPBDP_Form_Field_Type {
 				$checked = 'checked="checked"';
 			}
 
-			if ( $field->get_association() === 'meta' ) {
-				$html .= sprintf(
-					'<div class="%1$s"><label for="wpbdp-field-%5$d-%4$s"><input id="wpbdp-field-%5$d-%4$s" type="radio" name="%2$s" value="%4$s" %3$s /> %4$s</label></div>',
-					implode( ' ', $css_classes ),
-					'listingfields[' . $field->get_id() . ']',
-					$checked,
-					esc_attr( $label ),
-					$field->get_id()
-				);
-			} else {
-				$html .= sprintf(
-					'<div class="%1%s"><label for="wpbdp-field-%6$d-%5$s"><input id="wpbdp-field-%6$d-%5$s" type="radio" name="%2$s" value="%3$s" %4$s /> %5$s</label></div>',
-					implode( ' ', $css_classes ),
-					'listingfields[' . $field->get_id() . ']',
-					$option,
-					$checked,
-					esc_attr( $label ),
-					$field->get_id()
-				);
-			}
+			$html .= sprintf(
+				'<div class="%1$s"><label for="wpbdp-field-%6$d-%5$s"><input id="wpbdp-field-%6$d-%5$s" type="radio" name="%2$s" value="%3$s" %4$s /> %5$s</label></div>',
+				implode( ' ', $css_classes ),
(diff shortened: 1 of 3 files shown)


includes/fields/class-fieldtypes-radiobutton.php
```diff
@@ -72,14 +72,27 @@ public function render_field_inner( &$field, $value, $context, &$extra = null, $
...
-				'<div class="%1$s"><label for="wpbdp-field-%5$d-%4$s"><input id="wpbdp-field-%5$d-%4$s" type="radio" name="%2$s" value="%4$s" %3$s /> %4$s</label></div>',
-				implode( ' ', $css_classes ),
-				'listingfields[' . $field->get_id() . ']',
-				$checked,
-				esc_attr( $label ),
-				$field->get_id()
-			);
+			if ( $field->get_association() === 'meta' ) {
+				$html .= sprintf(
+					'<div class="%1$s"><label for="wpbdp-field-%5$d-%4$s"><input id="wpbdp-field-%5$d-%4$s" type="radio" name="%2$s" value="%4$s" %3$s /> %4$s</label></div>',
```
shervElmi, line 77: I’ve just tried rewriting this condition, and it seems possible. It could look something like the example below. I haven’t tested it yet, but here’s more context:

```php
$html .= sprintf(
    '<div class="%1$s"><label for="wpbdp-field-%6$d-%5$s"><input id="wpbdp-field-%6$d-%5$s" type="radio" name="%2$s" value="%3$s" %4$s /> %5$s</label></div>',
    implode(' ', $css_classes),
    'listingfields[' . $field->get_id() . ']',
    $field->get_association() === 'meta' ? esc_attr($label) : $option,
    $checked,
    esc_attr($label),
    $field->get_id()
);
```

includes/fields/class-fieldtypes-select.php
```diff
@@ -251,23 +251,35 @@ private function add_options_to_field( $options, $value, $field ) {
 			$option_data = array(
 				'label'      => $label,
 				'attributes' => array(),
+				'value'      => esc_attr( $option ),
```
shervElmi, line 254: Based on the WordPress Handbook (https://developer.wordpress.org/apis/security/escaping/#toc_3), we should always escape late. So, we don’t need `esc_attr` here.

I’ve checked, and we’re already escaping this value correctly later. We just need to remove the `esc_attr` here.

shervElmi (approved):
Thanks @sorinmarta, it looks great to me. 🚀

includes/fields/class-fieldtypes-select.php
```diff
@@ -251,23 +251,35 @@ private function add_options_to_field( $options, $value, $field ) {
...
 
 			$option_data = apply_filters( 'wpbdp_form_field_select_option', $option_data, $field );
 
-			$html .= sprintf(
-				'<option value="%3$s" class="%1$s" %2$s>%3$s</option>',
-				esc_attr( 'wpbdp-inner-field-option wpbdp-inner-field-option-' . WPBDP_Form_Field_Type::normalize_name( $option_data['label'] ) ),
-				self::html_attributes( $option_data['attributes'], array( 'value', 'class' ) ),
-				esc_html( $option_data['label'] )
-			);
+			if ( $this->is_multiple() && $field->get_association() == 'meta' ) {
```
shervElmi, line 267: It seems this condition could be managed without using if and else.

shervElmi: I mean like: https://github.com/Strategy11/business-directory-plugin/pull/431#discussion_r1930895065

sorinmarta:
Thanks @shervElmi!

Those were all great suggestions!

PR #432, opened 2025-01-28
Fix import button

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/285

This PR checks if the page has the import file inputs before adding the event listeners


shervElmi (approved):
Thanks @sorinmarta, it looks great to me! 🚀

PR #437, opened 2025-02-27
Add nonce and author check to ajax image upload

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/293

This PR adds nonce and author validation to listing image upload. It also adds a check to make sure the listing id is valid.


shervElmi (approved):
Thanks @sorinmarta, it looks good to me! 🚀

PR #440, opened 2025-04-11
Add placeholder option for required dropdowns

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/294#issuecomment-2794976259

This PR adds a default empty option to the required dropdown fields that don't have a `show_empty_option` property to ensure that it doesn't render the default option on the category field or any other custom dropdowns that have the empty option disabled.


shervElmi (approved):
Looks great to me. thank you @sorinmarta! 🚀

PR #443, opened 2025-04-23
Check for missing fields before creating defaults

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/262

This PR adds a check in the `create_default_fields` method that first checks if there are any missing required fields before creating the default ones to avoid duplications.


shervElmi (approved):
Looks great to me, Thanks @sorinmarta! 🚀 ❤️

PR #450, opened 2025-08-13
Add early check for content

sorinmarta:
fixes https://github.com/Strategy11/business-directory-premium/issues/311

This PR adds an early content check in the `wpbdp_has_shortcode` function to return early if `$content` is empty.


sorinmarta:
Thanks @shervElmi!

sorinmarta:
We've released it today @garretlaxton

PR #495, opened 2026-03-17
Fix issue in Stripe event handling

sorinmarta:
Check if the event needs to be handled before processing the event.


Crabcyborg (approved):
Looks good to me @sorinmarta!

Thanks for working on that!

🚀

